PR #47, opened 2014-01-08
Add simple version command

catsby:
This adds a very simple version command.
Usage:

```console
$ godep version
godep version: dev
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -56,6 +56,7 @@ var commands = []*Command{
 	cmdGet,
 	cmdPath,
 	cmdRestore,
+	cmdVersion,
 }
 
 func main() {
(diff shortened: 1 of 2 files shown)


shurcooL:
Just leaving a comment here as I've witnessed yet another instance where not having a version command caused problems for people. The person had the latest source of godep, but the installed binary was out of date and for some reason `go get -u github.com/tools/godep` did not update the binary (until it was removed and installed again).

Also, it's possible to set blank strings in Go via a linker flag, see http://technosophos.com/2014/06/11/compile-time-string-in-go.html article. Unfortunately, that will not be acceptable, because it requires the flag to be provided when building the binary, so it's not compatible with users doing `go get -u github.com/tools/godep`.

PR #48, opened 2014-01-17
Add a -extonly option to "godep save"

pedantic-git:
I have a project that includes Go components alongside other components, all organized together in the same repo.

By necessity, this project doesn't follow the directory naming conventions for Go packages, but its dependencies do.

My project's structure is:

```
/opt/myproj/src/packagename/packagename.go
/opt/myproj/python/...
```

where `/opt/myproj` is the root of the repository.

I'd like to be able to use godep for this, but unfortunately it makes a few assumptions that are incompatible with my project. The two assumptions that don't hold true for my project are:

1. `go list -e -json` always returns a "Root" parameter. It does not for my project when pwd = `/opt/myproj/src`
2. `p.Dir` and `p.Root` are never the same directory. If they are, you get this error:

```
godep: directory "/opt/myproj" is not using a known version control system
```

which looks to be an artifact of the [implementation of FromDir in vcs.go](https://code.google.com/p/go/source/browse/go/vcs/vcs.go?repo=tools#347).

I've added a "-extonly" option which resolves this for me, and thought you might be interested in taking it upstream.

It keeps track of the "homes" of the requested packages (their `p.Root` or `p.Dir` if there is no `p.Root`) and excludes packages found in that "home" from the dependency list.

I appreciate this is all pretty new to me and you may have a better solution to this problem. But here's a work-in-progress in case it provides some part of the overall solution to this problem.

### DIFF
--- a/dep.go
+++ b/dep.go
@@ -43,9 +43,9 @@ type Dependency struct {
 }
 
 // pkgs is the list of packages to read dependencies
-func (g *Godeps) Load(pkgs []*Package) error {
+func (g *Godeps) Load(pkgs []*Package, extonly bool) error {
 	var err1 error
-	var path, seen []string
+	var path, seen, home []string
 	for _, p := range pkgs {
 		if p.Standard {
 			log.Println("ignoring stdlib package:", p.ImportPath)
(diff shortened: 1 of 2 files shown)


JoeBruner:
I'm running into this same problem.  Looks like this issue is dead in the water though??

freeformz:
For the purposes of cleanup I'm going to close this as it's obvious that we're not going to merge it as is. If someone would like to re-open this issue with an up-to-date PR (perhaps with a different implementation) I'd be happy to review and consider/discuss inclusion.

Thanks!

PR #89, opened 2014-06-04
allow godep save without complaining when bzr is used

rogpeppe:
This error seems spurious, and means it's not possible to do:

    godep save -copy=false
    godep restore

if the project has a bzr dependency.

This may well be an insufficient fix - something else may also
required, but hopefully this should get the ball rolling.

### DIFF
--- a/save.go
+++ b/save.go
@@ -102,11 +102,6 @@ func save(pkgs []string) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


freeformz:
I am closing this because -copy=false hasn't been supported for a whule (since 754ee6f4e0e5fc8d0ef2692fb239d15c7a09dd84).

Issue #71, opened 2014-03-30
Godep doesn't work when symlinks are used to map projects into go workspace


lox:
I keep my projects under `~/Projects` and then sym-link it into my go workspace under `~/.go/src/github.com/xyz/xyz`. This doesn't seem to work with godep, when I run it (with a cwd of `~/.go/src/github.com/xyz/xyz`) I get the following error:

``` bash
$ godep save
godep: directory "/Users/lachlan/Projects/xyz/xyz" is outside source root "."
godep: error loading packages
```

By comparison, `go list ./...` works fine. 

If I copy the directory into place, rather than sym-link it, the problem goes away.

PR #106, opened 2014-08-05
When an external command fails, say what the command was in the error

autarch:
This changes all the spots where external commands are run to include the command that was run in the error string.

I was working on this as part of diagnosing #71

It turns out that the problem in #71 isn't related to running external commands from godep code, but I figured this might be nice to have anyway.

### DIFF
--- a/dep.go
+++ b/dep.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"log"
 	"os"
-	"os/exec"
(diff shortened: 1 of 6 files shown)


freeformz:
We have a bunch of other verbose output these days. If your are still interested in discussing something like this please re-open with an updated PR.

Issue #43, opened 2013-12-13
feature request: support alternate project layouts


insasho:
godep holds a strong opinion about how the source tree of a project is managed. See https://groups.google.com/forum/#!topic/golang-nuts/8NJq6jTIpas. Like Frank, our repository contains the entire source tree (and more). This allows ut do have a company-wide Makefile that implements various policies and procedures useful to our workflow, and allows stuff related to the project to be stored in the same git repo (ex: Vagrantfiles).  Also, we have a policy where all of our code must use the same dependencies.

In summary, this is our git repository structure:
```
/
/ops
/go
/go/src/company.com/proj1
/go/src/company.com/proj2
```

godep is incompatible with this structure. Example: `cd go/src/company.com && godep save` will fail with a "directory ../x/y/z/go is not using a known version control system", probably because /go is not the top-level directory of the repository.

godep has all of the other features we'd like in a dependency management system, but changing our model to support its requirements is too constraining for us. My request is that godep be adapted to support this pattern.

freeformz:
#108 is merged and may fix this.

PR #108, opened 2014-08-07
include command line packages in the set to copy

kr:
This is a simplification to how we define the behavior
of the save command. Now it has two distinct package
parameters, the "root set" and the "destination", and
they have clearer roles. The packages listed on the
command line form the root set; they and all their
dependencies will be copied into the Godeps directory.
Additionally, the destination (always ".") will form the
initial list of "seen" import paths to exclude from
copying.

In the common case, the root set is equal to the
destination, so the effective behavior doesn't change.
This is primarily just a simpler definition. However, if
the user specifies a package on the command line that
lives outside of . then that package will be copied.

As a side effect, there's a simplification to the way we
add packages to the initial "seen" set. Formerly, to
avoid copying dependencies unnecessarily, we would try
to find the root of the VCS repo for each package in the
root set, and mark the import path of the entire repo as
seen. This meant for a repo at path C, if destination
C/S imports C/T, we would not copy C/T into C/S/Godeps.
Now we don't treat the repo root specially, and as
mentioned above, the destination alone is considered
seen.

This also means we don't require listed packages to be
in VCS unless they're outside of the destination.

### DIFF
--- a/dep.go
+++ b/dep.go
@@ -42,9 +42,9 @@ type Dependency struct {
 }
 
 // pkgs is the list of packages to read dependencies
-func (g *Godeps) Load(pkgs []*Package) error {
+func (g *Godeps) Load(pkgs []*Package, destImportPath string) error {
 	var err1 error
-	var path, seen []string
+	var path []string
 	for _, p := range pkgs {
 		if p.Standard {
 			log.Println("ignoring stdlib package:", p.ImportPath)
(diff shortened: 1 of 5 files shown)


mikedoug:
Seems to also fix #48 as well?  Is there any way to help move things along with merging in this and other features people seem to be opening pull requests for?  I'm happy with this project as it solves a big problem with go dependencies; I just want to see it moving along faster to be more usable by people with real projects with go components that don't put the go source at the root of the repository.

mikedoug:
To ease the pain of getting this done, I've created a pull request (https://github.com/tools/godep/pull/173) to backmerge master into rootset.  That will update this branch to remove the merge conflict.

Then I'd like to help +1 and figure out how to champion getting the pull request approved.

freeformz:
I rebased this on top of master and fixed a problem with the diff command.

@kr @neurogeek I'm tempted to merge this at least because it stops a common issues: (a) where people are working in a repo that isn't in version control (yet) and (b) the go code that you want to vendor deps for is a subdirectory of the VCS project.

This would close #107, #108, #173 (And maybe #43 and #48).

Please review/comment.

Issue #84, opened 2014-05-15
save misses dependencies under build constraints


kr:
If a dependency is under build constraints that are not locally satisfied (for example, a foo_windows.go file on a unix machine), then godep save won't pick it up.

jzelinskie:
I believe that I created a duplicate issue at #161. However, some new information has come to light: if a library is using the `!` operator in a build line (i.e. `// +build !linux`),  even the `-tags` parameter on `go list` won't help.

freeformz:
Closing in favor of #271.

PR #117, opened 2014-08-31
godep save: add -tags flag.

jnfeinstein:
I needed this for a project and couldn't find any previous efforts, so I attempted it myself.  I based this on the fact that `go list` seemed to work fine with the `-tags` flag.  I ran it with my project and godep produced the correct outputs.  Please let me know if this is useful and/or done correctly.

### DIFF
--- a/pkg.go
+++ b/pkg.go
@@ -32,10 +32,19 @@ type Package struct {
 // Unlike the go tool, an empty argument list is treated as
 // an empty list; "." must be given explicitly if desired.
 func LoadPackages(name ...string) (a []*Package, err error) {
+	args := []string{"list", "-e", "-json"}
+	return runListCommand(args, name...)
+}
+
+func LoadPackagesWithTags(tags string, name ...string) (a []*Package, err error) {
+	args := []string{"list", "-e", "-json", "-tags", tags}
+	return runListCommand(args, name...)
+}
+
+func runListCommand(args []string, name ...string) (a []*Package, err error) {
 	if len(name) == 0 {
 		return nil, nil
 	}
-	args := []string{"list", "-e", "-json"}
 	cmd := exec.Command("go", append(args, name...)...)
 	r, err := cmd.StdoutPipe()
 	if err != nil {
--- a/save.go
+++ b/save.go
@@ -17,7 +17,7 @@ import (
 )
 
 var cmdSave = &Command{
-	Usage: "save [-r] [packages]",
+	Usage: "save [-r] [-tags=\"\"] [packages]",
 	Short: "list and copy dependencies into Godeps",
 	Long: `
 Save writes a list of the dependencies of the named packages along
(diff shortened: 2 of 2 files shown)


maxpersson:
Will this be merged? I just added tag based configuration for the storage of my CQRS library https://github.com/looplab/eventhorizon, without the proper tag on save dependencies for the specified storage backend are missing.

whazzmaster:
I'm also interested in this PR; I'd much rather use an official release rather than to fork it and apply the patch just to run goose on heroku.

jsha:
For the Let's Encrypt project (letsencrypt.org), we depend on tags to generate the correct dependency graph for godep, so the main branch doesn't work for us. The pull request fixes a critical issue for us, so currently we have to tell developers to fetch and install this branch instead of the default godep. It would be really helpful if you could merge this, @kr!

Thanks,
Jacob

freeformz:
@jnfeinstein Please rebase against master and I'll take a look.

jnfeinstein:
@freeformz, the branch is rebased.  The only conflict was in the usage blurb.

leehambley:
We just put an app into production, and we ran into this, we have conditional build-time metrics and analytics that rely on an external package (Rollbar), hidden behind builds tags, `godep` can't see it.

We've worked around it by importing the Rollbar package to `_` in a `// +build ignore` version of the file, which is less than ideal, since we still have to manually rewrite the import path in our `// +build release` file.

freeformz:
Thanks @jnfeinstein. At this point I think we need to have the new exported method documented and a test case.

kr:
I have hope that this can be solved without exposing additional knobs. The general approach I would take: godep should find all dependencies, ignoring `// +build` lines entirely. It should behave as if all build constraints are satisfied when finding imports. #84 has a little bit of discussion on this topic.

freeformz:
@kr That probably makes the most sense in most circumstances.

although I do wonder if there is a case where the developer doesn't support building everything with a dep that uses a different build tag then the one they designed for.

kr:
This approach wouldn't change what gets built, only what code is included in the repository.

freeformz:
That is correct, but since it's vendoring everything regardless of `// +build` it could include something the developer didn't intent to include/support, which could then be triggered with a `-tag` by the builder.

Not sure if it matters though.

kr:
> could then be triggered with a `-tag` by the builder

If that's the case, then the alternative is it wouldn't build at all for that builder.

It's still the responsibility of the person running `godep save` to at least glance at the diff to make sure it's reasonable, regardless of what godep does about tags.

kr:
I suppose not building at can be preferable to including unsupported code for an unsupported build configuration. Hmmm.

freeformz:
@jsha would it work for you guys if godep just included all the deps, ignoring `// +build` tags? Thereby punting the tagging off to those building the code later?

IMO this is the best solution.

jsha:
@freeformz: That sounds like the best solution to me as well, thanks!

freeformz:
Great I am going to close this then in favor of: #271

PR #119, opened 2014-09-05
Add an 'outdated' command to compare dependencies against what is in use

ivey:


neurogeek:
godep diff command has been merged to master and it should be able to do whats described here.

PR #128, opened 2014-10-10
Added support for build tags

nielsdraaisma:
Attempt to fix https://github.com/tools/godep/issues/84 , adds build tags support to all commands.


freeformz:
Thanks, but I think the way we're going to approach this by vendoring everything regardless of build tags.

PR #130, opened 2014-10-12
Support local (relative) imports

robert-zaremba:
Local imports are handled using "_/" prefix by a go tool.
Unfortunately this package didn't support local imports, which
are quiet popular in comercial products.


freeformz:
Thanks, but while this is supported by go it's actively discouraged and we won't support this.

PR #132, opened 2014-10-13
feature: added verbose flag to restore command

mariokostelac:
Restoring a big number of dependencies lasts for some time so it is nice to have some update about the progress.

### DIFF
--- a/restore.go
+++ b/restore.go
@@ -7,12 +7,20 @@ import (
 )
 
 var cmdRestore = &Command{
-	Usage: "restore",
+	Usage: "restore [-v]",
 	Short: "check out listed dependency versions in GOPATH",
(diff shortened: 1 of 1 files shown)


freeformz:
@mariokostelac Can you rebase this and I'll check it out?

freeformz:
#152 is similar(ish) and merged. If you want to re-open a PR with cleaned up (or additional) output that would be great, but I'm going to close this for now.

kr:
Did you mean #243? Looks like #153 is still open (but should be closed?).

freeformz:
@kr Yeah. I did. I think I meant to reference both #152 and #243, but got distracted part of the way through. Thanks

PR #139, opened 2014-10-23
godep save: Add interactive `-p` option

proullon:
With some projects splitted in several repositories, it comes handy to be able to select dependencies manually.

Inspired by `git add` command.
See : http://git-scm.com/docs/git-add

### DIFF
--- a/dep.go
+++ b/dep.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -52,6 +53,7 @@ type Dependency struct {
 func (g *Godeps) Load(pkgs []*Package) error {
 	var err1 error
 	var path, seen []string
+	reader := bufio.NewReader(os.Stdin)
 	for _, p := range pkgs {
 		if p.Standard {
 			log.Println("ignoring stdlib package:", p.ImportPath)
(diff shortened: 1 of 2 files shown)


dep.go
```diff
@@ -131,6 +133,13 @@ func (g *Godeps) Load(pkgs []*Package) error {
 			err1 = errors.New("error loading dependencies")
 			continue
 		}
+		if saveP {
+			fmt.Printf("godep: Would you like to index %s ?\n[Y/n]:", pkg.ImportPath)
+			text, _ := reader.ReadString('\n')
+			if text[0] == 'n' {
+				continue
+			}
```
alexandercampbell: This should be

```go
if len(text) < 1 || text[0] != 'y' {
        continue
}
```

Prevent out-of-bounds index panic and be a little stricter about the input.

dep.go
```diff
@@ -131,6 +133,13 @@ func (g *Godeps) Load(pkgs []*Package) error {
 			err1 = errors.New("error loading dependencies")
 			continue
 		}
+		if saveP {
+			fmt.Printf("godep: Would you like to index %s ?\n[Y/n]:", pkg.ImportPath)
```
alexandercampbell: To fit better with Unix convention this should fit on one line.

```go
fmt.Printf(`godep: index %q? [Y/n] `, pkg.ImportPath)
```

freeformz:
I don't understand the use case for selecting dependencies manually. The goal of godeps is to make builds repeatable by vendoring all dependencies. This seems to contradict this goal. 

Since this has been here a while I'm going to close this. Feel free to re-open with comments / mergable update if you still want to pursue this.

Thanks!

PR #143, opened 2014-11-04
Adding Godeps dependencies to godeps, tool should be maintaining its own dependencies

fkautz:


freeformz:
Already done in 3d123490b0a248b2dea64deee887504a63295717 so closing. Thanks!

PR #144, opened 2014-11-06
Copy tracked files only

### LINKED ISSUE #102: Untracked files are vendored from git repo without warning
If package `A` references git package `B`, and `B` has an untracked file `b.go`, running `godep save` in `A` will vendor `b.go` from `B` without a warning.

edsrzf:
This pull request adds a new VCS command to list all tracked files in a repository. When copying files, godep uses that list to make sure only tracked files are copied.

Closes #102

### DIFF
--- a/save.go
+++ b/save.go
@@ -280,9 +280,10 @@ func copySrc(dir string, deps []Dependency) error {
 			log.Println(err)
 			ok = false
 		}
+		files := dep.vcs.listFiles(dep.dir)
 		w := fs.Walk(dep.dir)
 		for w.Step() {
-			err = copyPkgFile(dir, srcdir, w)
+			err = copyPkgFile(files, dir, srcdir, w)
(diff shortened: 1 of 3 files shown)


dim:
+1. this would be a really nice flag as you currently may end up copying many unnecessary files. while these files will still be ignored, godep save can still take ages to run, because of a 100G test file, not mentioning the unnecessary bloat

freeformz:
@edsrzf Can you please rebase this and I'll take a look more?

edsrzf:
Done. I'll follow up with any minor requests, but if anything extensive is required, I'll probably just abandon this PR since we're no longer using godep.

It's been a while, but I think I only tested this with git too.

freeformz:
@edsrzf Thanks I changed a little and merged it. Also added a `-v` flag to go so you can see what's skipped if you want to.

PR #153, opened 2014-11-30
Add doc on `godep save -r ./...`

jingweno:


rafaeljusto:
+1

dmitris:
lack of documentaiton on "godoc save -r" creates confusion among users - see https://go-review.googlesource.com/#/c/8969/ for an example (review comment: "As I understand it, godep is a not a vendoring tool in the sense described in the faq (although it copies the source code, it does so by manipulating $GOPATH, not by rewriting import paths).")

freeformz:
I think I've effectively closed this with ba35b97d0db08d8d57599079d99b65add0282b4e.

Please let me know (new PR / Issue) if we should do anything more.

Thanks!

Issue #21, opened 2013-11-06
command: 'godep run'


kr:
https://groups.google.com/d/msg/go-package-management/-rNvdrdPC-k/ofin3XevusYJ

@davecheney:

> godep shell anyone?

@ardan-bkennedy:

> That is an awesome idea. That could really help make things easy for everyone. People don't want to think, "just tell me what to do". 

>The shell could identify the Godep folder(s) within a workspace, automatically setup the GOPATH and everything would run. It could also intercept calls and perform magic when necessary. All the go tooling works without needing to type godep first. It could handle some of the harder problems I am sure.

interlock:
:+1:

freeformz:
For now you should rewrite your deps so that you only need to use godep when managing dependencies. Long term it's likely the need to rewrite will go away pending the [vendoring experiment](https://go-review.googlesource.com/#/c/10923/).

PR #163, opened 2014-12-24
Add exec command

tarrant:
Exec will run a command and its args with the GOPATH set via godep.

Fixes https://github.com/tools/godep/issues/21

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -1,0 +1,33 @@
+package main
+
+import (
+	"log"
+	"os"
+	"os/exec"
+)
+
+var cmdExec = &Command{
+	Usage: "exec command [args]",
+	Short: "run a command with GOPATH set",
+	Long: `
+Exec sets the GOPATH to the vendored directory and then runs the specified command.
+`,
+	Run: runExec,
+}
+
+func runExec(cmd *Command, args []string) {
+	if len(args) == 0 {
+		log.Fatalln("Must specify command to execute")
+	}
+
+	c := exec.Command(args[0], args[1:]...)
+	c.Stdout = os.Stdout
+	c.Stderr = os.Stderr
+	c.Stdin = os.Stdin
+	c.Env = append(envNoGopath(), "GOPATH="+prepareGopath())
+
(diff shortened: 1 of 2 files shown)


jmervine:
+1

tarrant:
Hey @kr are there any fixes I can do to get this merged?

ready4god2513:
+1 here.  Anything holding it back?

jrdnull:
:+1:

freeformz:
`godep save -r ./...`, then you don't need `godep exec`

conslo:
`GOPATH=$(godep path) $COMMAND`

I don't feel like this is something this tool should be responsible for.

freeformz:
I don't think we're going to be adding an `exec` command in light of using rewriting now and the upcoming `-vendor` option to the go tooling. With either of those you should just be able to use the go toolchain normally and use godep "just" for managing dependencies.

I'm not sure if @kr agrees with me or not however.

ready4god2513:
Sounds good to me.  Thanks.

-- 
Brandon Hansen
Sent with Airmail

On June 17, 2015 at 11:29:34 AM, Edward Muller (notifications@github.com) wrote:

I don't think we're going to be adding an exec command in light of using rewriting now and the upcoming -vendor option to the go tooling. With either of those you should just be able to use the go toolchain normally and use godep "just" for managing dependencies.

I'm not sure if @kr agrees with me or not however.

kr:
I agree with @freeformz on this.

freeformz:
@tarrant Thanks for the PR, but I'm going to close this out, please consider using rewriting for now until go1.5's `-vendor` option becomes available.

conslo:
@freeformz could you point me in the right direction for information about this `-vendor` option? I did some googling but I seem to be failing to find anything.

freeformz:
@conslo Read down to the bottom: https://groups.google.com/forum/#!msg/golang-dev/74zjMON9glU/4lWCRDCRZg0J

Issue #140, opened 2014-10-28
godep does not skip "testdata" folders when enumerating Go packages.


shurcooL:
See the relevant code in [save.go#L293](https://github.com/tools/godep/blob/754ee6f4e0e5fc8d0ef2692fb239d15c7a09dd84/save.go#L293):

```Go
if c := w.Stat().Name()[0]; c == '.' || c == '_' {
	// Skip directories using a rule similar to how
	// the go tool enumerates packages.
	// See $GOROOT/src/cmd/go/main.go:/matchPackagesInFs
	w.SkipDir()
}
```

But `$GOROOT/src/cmd/go/main.go:/matchPackagesInFs` in Go 1.3.3 has:

```Go
// Avoid .foo, _foo, and testdata directory trees, but do not avoid "." or "..".
_, elem := filepath.Split(path)
dot := strings.HasPrefix(elem, ".") && elem != "." && elem != ".."
if dot || strings.HasPrefix(elem, "_") || elem == "testdata" {
	return filepath.SkipDir
}
```

Notably, it skips the "testdata" folder. `godep` should do the same. The "testdata" folder is sometimes used to have invalid .go code files (e.g., for testing the Go parser) and other such things inside, which may prevent otherwise valid Go packages from working with `godep`.

The behavior of `go` tool skipping "testdata" folders has been recently documented, see https://codereview.appspot.com/148170043.

shurcooL:
This issue prevents godep from working in some cases:

```bash
$ godep save
godep: Godeps/_workspace/src/golang.org/x/tools/go/loader/testdata/badpkgdecl.go:1:34: expected 'package', found 'EOF'
```

It should not be reading .go files inside testdata folder!

shurcooL:
There's a PR #166 that fixes it, can it be merged please?

shurcooL:
There's a PR #166 that fixes this issue, can it be reviewed and merged please?

neurogeek:
I just pushed a fix to ignore testdata directories. Thanks @shurcooL for bringing this up!

shurcooL:
For reference, this issue was resolved in commit bb418303c7e0886d29a75b5bbec49b64cdc3fa1c, not closed as "wontfix".

(As a side note, if you had used one of the [supported keywords](https://help.github.com/articles/closing-issues-via-commit-messages/#keywords-for-closing-issues) like "resolves", "fixes" or "closes", that would've been clearer. "addresses" is not one of the supported keywords.)

[...]

PR #166, opened 2014-12-29
fix to skip 'testdata' directory #140

atomaths:


shurcooL:
LGTM because it fixes the issue, but I would copy and paste the way the Go tool does this exactly, instead of using a different syntax. That way it's easier to keep it in sync with any future changes to the Go tool.

```Go
// Avoid .foo, _foo, and testdata directory trees, but do not avoid "." or "..".
elem := w.Stat().Name()
dot := strings.HasPrefix(elem, ".") && elem != "." && elem != ".."
if dot || strings.HasPrefix(elem, "_") || elem == "testdata" {
```

Source: https://github.com/golang/go/blob/dc870d5f4b77c0a2a276f8fd12510a53969c56b5/src/cmd/go/main.go#L616-L619

shurcooL:
Actually, this may not fix issue, looking into it.

shurcooL:
Okay, I've done some testing regarding https://github.com/tools/godep/issues/140#issuecomment-98222302, and this fix does work. But the problem is that it's not retroactive. Workspaces saved with older godep will need to be removed and recreated using new version, which will skip over testdata folder and avoid the issue in the future.

shurcooL:
Actually, I will retract my LGTM and say -1, I'd prefer this not be merged as is.

I think what it fixes is a symptom, not the real cause. It should be okay to vendor `testdata` folders, the issue is that they shouldn't be navigated into when listing Go packages and causing errors since `testdata` folders may purposefully (correctly) contain .go files that are invalid Go code.

That issue should  be fixed first, then perhaps `testdata` folders should be skipped when vendoring also. After all, folders starting with `.` and `_` are already skipped.

I'll try to look into this more closely in some time.

PR #167, opened 2014-12-31
save.go: missing f.Close()

gprggr:


kr:
Thanks for the fix! :grinning:

PR #169, opened 2015-01-07
Don't copy GIT_ env variables so we can run godep in a githook

ErikDubbelboer:
When running `git` we should not copy all environment variables. When `godep` is run from a `githook` git will set some environment variables like `GIT_DIR` and `GIT_INDEX_FILE`. These variables end up being set for the `git` command run by `godep` which prevents it from working correctly. `git` seems to prefer the variables set in the environment before `--git-dir=` and will somehow complain that the directory is not a repository.

I know that removing the environment variables in `run1` is a bit nasty seeing that this function is also being used to run commands for other vcs's. I could add a field in the `VCS` which sets the prefix of environment variables to be removed. I'm not sure if that is a more clean and future proof.

### DIFF
--- a/vcs.go
+++ b/vcs.go
@@ -194,6 +194,12 @@ func (v *VCS) run1(dir string, cmdline string, kv []string, verbose bool) ([]byt
 	var buf bytes.Buffer
 	cmd.Stdout = &buf
 	cmd.Stderr = &buf
+	cmd.Env = make([]string, 0)
+	for _, kv := range os.Environ() {
+		if !strings.HasPrefix(kv, "GIT_") {
+			cmd.Env = append(cmd.Env, kv)
+		}
+	}
 	err = cmd.Run()
 	out := buf.Bytes()
 	if err != nil {


conslo:
this problem makes sense, but instead of a blacklist which gets things working via a curated magic list of "stuff that breaks other commands", maybe it should be a whitelist of "this is what we need to do what *we* do". That way instead of being responsible for finding every environment variable that might break anything, the tools is only responsible for knowing what environment variables it needs to do *its* thing(s), and it can be established as a rule that "commands this runs are isolated from the environment".

This idea may also prompt some sort of flag to specify other environment variables to pass through, should someone's need arise to break this isolation.

freeformz:
Thank, but ...

I don't think godep should be messing around with GIT_* environment variables. For instance this would mess with my own setup in a negative way as I rely on GIT_AUTHOR_EMAIL & GIT_AUTHOR_NAME at times. If you are using godep inside of a git hook you should unset GIT_DIR/GIT_INDEX_FILE before calling godep.

FWIW: I'd be happy to re-visit this if a different solution is presented.

ErikDubbelboer:
I don't really understand how you are using GIT_AUTHOR_EMAIL & GIT_AUTHOR_NAME inside godep. But I agree with closing the issue as it's easy to unset the variables in the hook itself.

freeformz:
@ErikDubbelboer Those env vars are not being used inside of godep, but I am using them in my environment. Thanks for contributing.

PR #173, opened 2015-01-15
Rootset

mikedoug:
This performs a backmerge of the current state of master into this aging rootset branch.  I've verified the changes are consistent with the original rootset branch, and we've tested it in a project in which we needed this additional functionality.

My goal is to get this backmerge done, and then lobby for merging tools/godep/rootset into tools/godep/master such that this is codified as supported.  This appears to solve #43 #48 and #108.


freeformz:
#108 is merged, so closing.

Issue #103, opened 2014-07-16
Import error for Google Appengine packages


matthewbelisle-wf:
Google Appengine is a common platform for Go apps, but godep can't save appengine packages because the appengine SDK shims its import paths instead of using normal VCS based import paths.  So instead of `import "github.com/something/appengine/datastore"` it does `import "appengine/datastore"` and then shims the imports somehow to point to the SDK paths for the package.

https://github.com/GoogleCloudPlatform/appengine-guestbook-go/blob/32112ced8de3e8e02bd64801f99f776527c62f23/hello.go#L8

This is a problem because godep throws an error when it tries to import these,

```
$ godep save .
godep: cannot find package "appengine/datastore" in any of:
        /usr/local/Cellar/go/1.3/libexec/src/pkg/appengine/datastore (from $GOROOT)
        /Users/mgbelisle/go/src/appengine/datastore (from $GOPATH)
godep: error loading dependencies
```

which makes sense but is unfortunate.  What is the proper way to fix that?  This change here works but seems a little kludgy.

https://github.com/matthewbelisle-wf/godep/commit/dba190f14fc83759b74df44e4ab4e7a492eedb1d

fredr:
+1

fredr:
@kr the -ignore flag, do you want it to be a comma separated list of import paths, or a regex?

arschles:
+1 on this. I'd be ok with CRLF delimited

vvakame:
:+1:

sinmetal:
:+1:

freeformz:
@neurogeek you may be interested in this one?

neurogeek:
Sure, I'll take a look later today.

kalbasit:
I don't believe that this is necessary any longer as the legacy appengine packages were replaced by the ones that run on both appengine and managed appengine. See [this](https://code.google.com/p/googleappengine/issues/detail?id=11670) and [this](https://github.com/golang/appengine#updating-a-go-app-engine-app) for more information.

I am currently using [`google.golang.org/appengine`](https://github.com/go-acd/oauth-server/blob/master/goapp/roothandler.go#L10) in my project hosted by Google Appengine (not within a Managed VM). I just omitted `vm:true` from [`app.yaml`](https://github.com/go-acd/oauth-server/blob/master/appengine/app.yaml).

neurogeek:
@kalbasit: Interesting. With new qualification over appengine packages, we definitely don't need this.
[...]

PR #175, opened 2015-01-20
add default omit patterns, and a omit cli flag

fredr:
Proposed fix for #103 

Omit appengine and appengine_internal by default, and adds a omit
flag for other package patterns.


freeformz:
Thanks, but In #103 it was decided that people should upgrade their apps to the current appending libraries which aren't so strange AFAICT.

PR #182, opened 2015-02-12
rewrite: remove existing file before moving the temp file into place.

kardianos:
Windows doesn't allow a rename over an existing file.

The other way to fix this problem on windows is to do a three step rename old (".old"), rename new, then delete old. However, I think this is fine.


jefff:
:+1:

freeformz:
This was fixed in https://github.com/tools/godep/commit/6a67fc143981c5d893576c3fbeb964a3745d7b12

Thanks for taking the time to submit this.

PR #188, opened 2015-03-02
Support for resolving symbolic links 

nl5887:
This should fix issue #71. It will resolve eventual symbolic links and project them on the source folder. This allows to create symbolic links from several projects to the gopath folder.

I've seen the discussion around this, and if you still prefer **won't fix** just ignore and close this PR ;-)

### DIFF
--- a/save.go
+++ b/save.go
@@ -276,10 +276,30 @@ func copySrc(dir string, deps []Dependency) error {
 			log.Println(err)
 			ok = false
 		}
-		w := fs.Walk(dep.dir)
+
+		var walkDir string
+		if walkDir, err = filepath.EvalSymlinks(dep.dir); err != nil {
+			return err
+		}
+
+		w := fs.Walk(walkDir)
 		for w.Step() {
-			err = copyPkgFile(dir, srcdir, w)
+			// project the relative (symlink) dir to the src dir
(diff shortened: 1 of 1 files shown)


nickckaye:
@kr I prefer to version control my entire platform at once, however Golang is just one of the two modules. So my only solution at the moment is to keep my whole project in the `~/Go/` folder. Not only is this confusing, it forces me to create new proprietary workflows just for my other langs to play well with Go! Bullyish..

vikstrous:
This would be really nice. If this patch works, I think it should be merged. IMO it doesn't make sense to have an intentional bug because it's not the "go way".

freeformz:
The stance of the go team/tooling is "don't use symlinks". Godep ends up utilizing that go tools / libs underneath and I'm of the opinion that supporting this will cause more issues than it solves.

So thank you for spending time on this issue, but I'm not going to accept this patch atm.

charneykaye:
Thanks @freeformz . Just a comment for future readers- I ended up coming down on the side of "don't use symlinks" myself, and here's why.

Deployed into production, we can put our binaries absolutely anywhere we like. That's the beauty of the compiled language.

So the conversation about where to put our Go working files is really one about development workflow. And there are some very cool aspects of the Go open-source ecosystem that are facilitated by the federated directory structure.

PR #191, opened 2015-03-05
Follow symlinks one level down on save.

jmervine:
Hello,

The commit add support for following a symlink one level down when using `godep save`. I'm somewhat new to `golang`, so any pointers on a better way to handle this would be awesomeand I'll gladly take a stab at it. 

#### Why?

I've seen on several occasions where versioning is done via a symlink (example: github.com/jmervine/exec/v2) with a file structure something like:

```
# where v2 and latest are the current and symlinked path
├── exec.go
├── exec_test.go
├── latest -> .
├── README.md
├── _support
│   ├── slow.sh
│   └── test.sh
├── v1
│   ├── exec.go
│   ├── exec_test.go
│   ├── README.md
│   └── _support
│       ├── slow.sh
│       └── test.sh
└── v2 -> .
```

When using `godep save` with the context of `github.com/jmervine/exec/v2`, the expectation is a result of something like:

```
Godeps
├── Godeps.json
├── Readme
└── _workspace
    └── src
        └── github.com
            └── jmervine
                └── exec
                    └── v2
                         ├── exec.go
                         ├── exec_test.go
                         └── README.md
```

> Known issue: If the sub-package has a directory structure below it, it will not copy the resulting directories and files.

### DIFF
--- a/save.go
+++ b/save.go
@@ -301,6 +301,38 @@ func copyPkgFile(dstroot, srcroot string, w *fs.Walker) error {
 		// See $GOROOT/src/cmd/go/main.go:/matchPackagesInFs
 		w.SkipDir()
 	}
+
+	// follow symlinks one level deep
+	if w.Stat().Mode()&os.ModeSymlink == os.ModeSymlink {
+		fis, err := ioutil.ReadDir(w.Path())
+		if err != nil {
+			log.Println(err)
+		}
+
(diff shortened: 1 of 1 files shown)


jmervine:
@nl5887 and I must have been working on this at the same time, as #188 wasn't there when I first looked. His implementation is better. Cheers!

Issue #186, opened 2015-02-24
Godep breaks when go get fails to fetch a non-existing package in an existing repo


fkautz:
In the github.com/docker/distribution project, godep is setting an import package as:
```
github.com/MSOpenTech/azure-sdk-for-go/clients/storage
```
instead of 
```
github.com/MSOpenTech/azure-sdk-for-go
```

The result is 
```
 godep restore
package github.com/MSOpenTech/azure-sdk-for-go/clients/storage
	imports github.com/MSOpenTech/azure-sdk-for-go/clients/storage
	imports github.com/MSOpenTech/azure-sdk-for-go/clients/storage: cannot find package "github.com/MSOpenTech/azure-sdk-for-go/clients/storage" in any of:
	/home/fkautz/opt/go1.4.1/src/github.com/MSOpenTech/azure-sdk-for-go/clients/storage (from $GOROOT)
	/home/fkautz/gopath/src/github.com/MSOpenTech/azure-sdk-for-go/clients/storage (from $GOPATH)
```

fkautz:
Further analysis, godep uses go get.
```
github.com/MSOpenTech/azure-sdk-for-go/clients/storage
```
results in
```
go get github.com/MSOpenTech/azure-sdk-for-go/clients/storage
```
which fails because clients/storage is not in master.

However, the repo is properly downloaded. All that remains is the vcs revision being set to the correct version. (DL succeeds thanks to https://github.com/golang/go/blob/master/src/cmd/go/vcs.go#L697-L701)

One potential fix would be to detect "cannot find package" in the error, traverse to the repo location (as identified in vcs.go or ending in .vcs), and performing the vcs revision from the root of the repo. A go get without -u should result in a successful package download at that point to verify it worked.

Do you have any recommendations? I can help implement a fix for this.

kr:
This is reasonable for a start. Another possibility would be to copy code from the go tool to download all packages in process. This would give much more control over this process, and let us fix one or two other issues as well. However, that would be a lot more work.

For checking the error output of `go get`, there are a couple of things to consider:

1. If the `godep restore` is eventually successful, it should produce be no output.
2. If it is not successful, each error should be printed exactly once.

fkautz:
I am ok with copying code over. I'll put together an initial attempt.

fkautz:
@kr: Here is a working restore. I have given you access to github.com/fkautz/godep so we can collaborate on this more efficiently.

[...]

PR #196, opened 2015-03-24
Pulling in cmd/go and accessing it directly instead of via exec

fkautz:
So, in this patch, we pull in cmd/go, make go get private methods public, do a lot of tree shaking (we don't need all cmd/go), and make godep use those new public methods with some minor fixes (like not checking directory for buildable files before reverting to older commit). :)


fkautz:
@kr Does this patch make sense, based on what we spoke about before? If not, I can try submitting a different approach to fix #186.

fkautz:
Looks like this issue was fixed in go1.5 due to https://github.com/golang/go/issues/10509

Closing.

PR #201, opened 2015-04-01
Rewrite should ignore symbolic linked .go file

activars:
Hi,

We have use cases of keeping symbolic linked `.go` files in the project root directory. The current  rewrite logic modifies the symbolic link files when `godep save ./...` is executed.

This PR ensures the rewrite logic leave symbolic link file untouched.

Thank you

### DIFF
--- a/rewrite.go
+++ b/rewrite.go
@@ -48,11 +48,20 @@ func rewriteTree(path, qual string, paths []string) error {
 			log.Println("rewrite:", w.Err())
 			continue
 		}
-		if !w.Stat().IsDir() && strings.HasSuffix(w.Path(), ".go") {
-			err := rewriteGoFile(w.Path(), qual, paths)
+		filePath := w.Path()
+		if !w.Stat().IsDir() && strings.HasSuffix(filePath, ".go") {
+			fileInfo, err := os.Lstat(filePath)
 			if err != nil {
 				return err
 			}
+
+			if fileInfo.Mode() != os.ModeSymlink {
+				err = rewriteGoFile(filePath, qual, paths)
+				if err != nil {
+					return err
+				}
+			}
+
 		}
 	}
 	return nil


rewrite.go
```diff
@@ -48,11 +48,20 @@ func rewriteTree(path, qual string, paths []string) error {
 			log.Println("rewrite:", w.Err())
 			continue
 		}
-		if !w.Stat().IsDir() && strings.HasSuffix(w.Path(), ".go") {
-			err := rewriteGoFile(w.Path(), qual, paths)
+		filePath := w.Path()
```
neurogeek: s/filePath/path

rewrite.go
```diff
@@ -48,11 +48,20 @@ func rewriteTree(path, qual string, paths []string) error {
...
+		if !w.Stat().IsDir() && strings.HasSuffix(filePath, ".go") {
+			fileInfo, err := os.Lstat(filePath)
 			if err != nil {
 				return err
 			}
+
+			if fileInfo.Mode() != os.ModeSymlink {
```
neurogeek: s/fileInfo/info

I'd rather have something like:
info.Mode() == os.ModeSymlink {
  continue
}

err = rewriteGoFile ...

neurogeek:
I think this is reasonable. @freeformz what do you think?

freeformz:
@neurogeek +1 on your suggestions.

Although I am personally wary of adding this. If you don't rewrite the files, you'll end up picking up the same dep from elsewhere. AFAIK if you could get 2 copies of the dependencies initialized. I think this will add confusion overall.

neurogeek:
@freeformz Thanks for your input!.
@activars Could you expand on the rationale behind the symlinks? So I can better understand your use case?

Cheers,

activars:
We have multiple main binaries in the same project. We keep the main1.go main2.go in a `main` package. Then, symbolic link mains to the root directory for convenient building/dev purpose. It was frustrated Godep replace linking files in local development.

freeformz:
@activars 

Can't you use a directory structure like so:

```
./cmd/cmd-b-name/main.go
./cmd/cmd-b-name/main.go
./common_package.go
```

The above is installable with `go install ./...` or via `go get github.com/activars/mypackage/...` (at least with rewrites).

If you can't can you please expand on your explanation as to why, perhaps with an example?

freeformz:
Also, the stance of the go team/tooling is "don't use symlinks". Godep ends up utilizing that go tools / libs underneath and I'm of the opinion that supporting this will cause more issues than it solves. This is one example of them.

So thank you for spending time on this issue, but I'm not going to accept this patch atm.

PR #214, opened 2015-05-06
GFW golang.org

flyskywhy:
Hi,

Because of the nasty GFW from the Chinese government, golang.org can't be accessed from China especially in shell like "go get github.com/tools/godep", so I replace "golang.org/x/tools" with its mirror "github.com/golang/tools".

Would you please accept the pull request for this reason?

Thank you

### DIFF
--- a/vcs.go
+++ b/vcs.go
@@ -8,7 +8,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	"golang.org/x/tools/go/vcs"
+	"github.com/golang/tools/go/vcs"
 )
 
 type VCS struct {


cybertk:
+1

freeformz:
A workaround for this would be to checkout the github golang tools git repo to `$GOPATH/src/golang/x/tools`. After that you can use `import "golang.org/x/tools..."` and  `godep save -r ./...` as needed.

kr:
We could also solve this by using godep on itself. ;)

freeformz:
@kr Yes, perhaps we should.

PR #215, opened 2015-05-12
Skip paths that are not version controlled

dprotaso:
If godep now encounters a path that is not version controlled instead of quiting
and logging an error it will silently skip this path.

### DIFF
--- a/dep.go
+++ b/dep.go
@@ -55,12 +55,8 @@ func (g *Godeps) Load(pkgs []*Package) error {
 			err1 = errors.New("error loading packages")
 			continue
 		}
-		_, reporoot, err := VCSFromDir(p.Dir, filepath.Join(p.Root, "src"))
-		if err != nil {
-			log.Println(err)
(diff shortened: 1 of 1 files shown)


freeformz:
I don't know if this is a good idea. What's the use case?

dprotaso:
This loosens the requirement that every package you use needs have it's own .git/vcs root folder. This makes the tool more flexible to alternative source layouts in your vcs repo.

freeformz:
Wouldn't you want Godep to vendor the directory though?

dprotaso:
For my use case no - but it should.

I'll make the following additions:
- disable the vcs check with a flag (hence the default behaviour will remain the same)
- vendor all the directories

freeformz:
@dprotaso any updates?

freeformz:
I'm going to close this out for now since there haven't been updates in a while. Please feel free to re-open / start new PR if you still think this feature is valuable.

PR #217, opened 2015-05-15
Resolve vcs type the same way as 'go get -u' does.

alexvasi:
Basically it fixes `restore` command for private repositories.


freeformz:
Thanks!

Merged in 8063f550a728b76d3a1a828e45d482f7dbde61e7

PR #218, opened 2015-05-18
Comments are preserved when the imports are sorted

jeeyoungk:
the code cannot both modify the imports & run sorting at the same time. Without the fix, we get the following error:

```
~/Development/go/src/github.com/tools/godep (ddcfe27...) go test -v -run TestRewrite
=== RUN TestRewrite
--- FAIL: TestRewrite (0.02s)
	save_test.go:1068: godeptest/src/C/main.go = package main
		
		import (
			"C/Godeps/_workspace/src/D" // src E
			"C/Godeps/_workspace/src/E" // src D
		)
		 want package main
		
		import (
			"C/Godeps/_workspace/src/D" // src D
			"C/Godeps/_workspace/src/E" // src E
		)
```

The fix does two rounds parsing and saving, by `parse-rewrite-save-parse-sort-save` instead of `parse-rewrite-sort-save`.

### DIFF
--- a/rewrite.go
+++ b/rewrite.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"log"
 	"os"
 	"path/filepath"
@@ -61,6 +62,7 @@ func rewriteTree(path, qual string, paths []string) error {
 // rewriteGoFile rewrites import statments in the named file
 // according to the rules for func qualify.
 func rewriteGoFile(name, qual string, paths []string) error {
+	printerConfig := &printer.Config{Mode: printer.TabIndent | printer.UseSpaces, Tabwidth: 8}
 	fset := token.NewFileSet()
 	f, err := parser.ParseFile(fset, name, nil, parser.ParseComments)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


jeeyoungk:
CC: @kr

freeformz:
@jeeyoungk what happens to the extra intermediate file?

jeeyoungk:
@freeformz actually, that's a good point. i haven't though of cleaning it up. I should do that

kr:
The parser can accept a []byte or string as the third argument. It would be nice not to create a temporary file in the filesystem, but instead keep the intermediate file in memory.

jeeyoungk:
@kr didn't know about the other parameter. now it's using that. (no more intermediary temp file). I added a test to `glob` for any lingering temp files also.

freeformz:
@jeeyoungk you can just pass in the buffer itself as it also takes an io.Reader, which bytes.Buffer is.

jeeyoungk:
@freeformz thanks. didn't know about that. applied the changes.

rewrite.go
```diff
@@ -82,21 +84,25 @@ func rewriteGoFile(name, qual string, paths []string) error {
 	if !changed {
 		return nil
 	}
-
+	var buffer bytes.Buffer
+	if err = printerConfig.Fprint(&buffer, fset, f); err != nil {
+		return err
+	}
+	fset = token.NewFileSet()
+	f, err = parser.ParseFile(fset, "", &buffer, parser.ParseComments)
```
freeformz: I think you'd still want a unique fname here as the filename is still used to "record position information", but I'm not really 100% sure.

jeeyoungk: I changed to to use the original `name` again, but i'm not sure what effect here.

jeeyoungk:
Please take a look

freeformz:
LGTM. I tested this on a few random projects and it also fixes the bug originally reported.

jeeyoungk:
I've squashed the commits, so it's ready to be merged anytime.

kr:
LGTM

kr:
Thanks so much!

PR #219, opened 2015-05-19
Use a bytes.Buffer instead of an intermediate file

freeformz:
Extends the work of @jeeyoungk in #218


freeformz:
Closing as @jeeyoungk updated his PR with basically the same thing.

Issue #228, opened 2015-06-12
Vendoring latlong with godep


pshevtsov:
I'm trying to vendor [latlong] dependencies with Godeps. But when I run

```bash
$ godep save
```

I get the following error message:

```
godep: Godeps/_workspace/src/github.com/bradfitz/latlong/z_gen_tables.go:23:18: expected '}', found 'EOF'
```

[latlong]: https://github.com/bradfitz/latlong

freeformz:
Godep is truncating data, probably because the line being truncated is longer than 64k. If I'm right I may have a fix soon.

freeformz:
This should now be fixed by #231, so please `go get -u github.com/tools/godep` and re-open if this is still a problem (PS: You may need to cleanup your broken workspace first).

PR #179, opened 2015-02-01
Handle files with lines longer than 64kb

smarterclayton:
bufio.Scanner will return an error if a line is longer than 64kb bytes. This changes the import rewrite code to use the lower level bufio.Reader.ReadLine method, which returns the line in chunks.

Fixes #160

Issue #160, opened 2014-12-16
save and update strips out long lines of code


9uuso:
Hello,

I've faced really confusing error with packages which have really long source code lines. I'm not sure is this really an error in `godep` or is it a bug in Go standard library, but I thought it might be a good idea to start here.

Basically, I was trying to import package [`unicode`](https://github.com/rainycape/unidecode), which built fine with my application on my development machine. However, after dumping the dependency and trying to run tests for it on my CI service, I got error saying that the variable [`tableData`](https://github.com/rainycape/unidecode/blob/master/table.go#L5) was undefined.

I initially though this was a problem with the `unidecode` package's git import path, so I vendored the package, but it did not fix the problem.

Eventually I tried importing the variable in my [vendored package](https://github.com/9uuso/unidecode/blob/master/unidecode.go#L29) by splitting the line into five variables.

When I tried to include the updated package in my project, I noticed from that SourceTree indicated line removals after running `go get -u github.com/9uuso/unidecode` and `godep update github.com/9uuso/unidecode`. These line removals were made after [line 28](https://github.com/9uuso/unidecode/blob/master/unidecode.go#L28). Line 28 was now the last line of the file.

![screenshot 2014-12-16 21 17 10](https://cloud.githubusercontent.com/assets/1684211/5460132/f9020f58-8568-11e4-9c64-7a0898b98dd6.png)

I now checked my application's git history and saw that the line where the `tableData` should have been, [was missing](https://github.com/9uuso/vertigo/blob/f0730970e994ee45cc90ac2746df43bb9f576edf/Godeps/_workspace/src/gopkgs.com/unidecode.v1/table.go), even though when comparing with the original package, [it was there](https://github.com/rainycape/unidecode/blob/master/table.go#L5). The git commit hash in the [Godeps.json](https://github.com/9uuso/vertigo/blob/f0730970e994ee45cc90ac2746df43bb9f576edf/Godeps/Godeps.json#L160) and in the latest [unidecode commit](https://github.com/rainycape/unidecode/commit/4deae2c05236b41cc39f8144ac87a837ba974d40) matched.

The bug can be worked around by manually adding the line to the Godeps folder.

Any input on this?

felixge:
[...]

PR #231, opened 2015-06-17
Handle lines longer than 64k

freeformz:
Handle lines longer than 64k.

/cc @kr #228 #160 #179

### DIFF
--- a/save.go
+++ b/save.go
@@ -371,7 +371,6 @@ func copyWithoutImportComment(w io.Writer, r io.Reader) error {
 			}
 		}
 
-		// If we're at the EOF exit
 		if eof {
 			return nil
 		}
--- a/save.go
+++ b/save.go
@@ -354,26 +354,19 @@ func copyFile(dst, src string) error {
 }
 
 func copyWithoutImportComment(w io.Writer, r io.Reader) error {
-	var eof bool
 	b := bufio.NewReader(r)
 	for {
 		l, err := b.ReadBytes('\n')
-		if err != nil {
-			if err == io.EOF {
-				eof = true
-			} else {
-				return err
-			}
+		eof := err == io.EOF
+		if err != nil && err != io.EOF {
+			return err
 		}
 
 		// If we have data then write it out...
 		if len(l) > 0 {
-			// Trim the \n if it exists
-			if l[len(l)-1] == '\n' {
-				l = l[0 : len(l)-1]
-			}
-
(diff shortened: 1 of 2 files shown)


save.go
```diff
@@ -354,14 +354,35 @@ func copyFile(dst, src string) error {
...
-	sc := bufio.NewScanner(r)
-	for sc.Scan() {
-		_, err := w.Write(append(stripImportComment(sc.Bytes()), '\n'))
+	var eof bool
+	b := bufio.NewReader(r)
+	for {
+		l, err := b.ReadBytes('\n')
 		if err != nil {
-			return err
+			if err == io.EOF {
```
kr: It would be pretty cool to pull this out in front of the surrounding if stmt:

```go
l, err := b.ReadBytes('\n')
eof := err == io.EOF
if err != nil && err != io.EOF {
  return err
}
```

save.go
```diff
@@ -354,14 +354,35 @@ func copyFile(dst, src string) error {
...
+				eof = true
+			} else {
+				return err
+			}
+		}
+
+		// If we have data then write it out...
+		if len(l) > 0 {
+			// Trim the \n if it exists
+			if l[len(l)-1] == '\n' {
```
kr: I think bytes.TrimRight would be more readable here..

save.go
```diff
@@ -354,14 +354,35 @@ func copyFile(dst, src string) error {
...
+				l = l[0 : len(l)-1]
+			}
+
+			if _, err := w.Write(append(stripImportComment(l), '\n')); err != nil {
```
kr: Since the call to w.Write is made primarily for its side effect, IMO it doesn't belong inside the if-stmt, it should go on its own line above.

freeformz:
@kr Done. If it LGTY, then I'll merge it.

save.go
```diff
@@ -354,14 +354,28 @@ func copyFile(dst, src string) error {
...
+		// If we have data then write it out...
+		if len(l) > 0 {
+			// Strip off \n if it exists because stripImportComment
+			_, err := w.Write(append(stripImportComment(bytes.TrimRight(l, "\n")), '\n'))
+			if err != nil {
+				return err
+			}
+		}
+
+		// If we're at the EOF exit
```
kr: The comments in this function don't really add anything meaningful. I would delete them.

freeformz: Removed.

kr:
+1 LGTM

Issue #1, opened 2013-05-09
diff command


brianm:
It's useful to be able to see differences between the saved list of deps and the installed set of deps in the workspace without having to run `godep save` and without thereby clobbering the saved list.

Let's consider a command `godep diff`:

```
$ godep diff
--- Godeps
+++ $GOPATH
@@ -25,7 +25,7 @@
 		},
 		{
 			"ImportPath": "github.com/kr/fernet",
-			"Rev": "9596341a4fd02e93cd80509911ea9ed4cd84e7bf"
+			"Rev": "54c3c8fb9a0701721fceaa28780ce25ef447fd4f"
 		},
 		{
 			"ImportPath": "github.com/kr/sshpool",
```

or

```
$ godep diff
--- Godeps
+++ $GOPATH
@@ -24,6 +24,10 @@
 			"Rev": "b2f2731352d9db3b628d6ff086741de7530c10a9"
 		},
 		{
+			"ImportPath": "github.com/kr/fernet",
+			"Rev": "9596341a4fd02e93cd80509911ea9ed4cd84e7bf"
+		},
+		{
 			"ImportPath": "github.com/kr/sshpool",
 			"Rev": "f7f9e575cb90c1cb120028cc228b4e458053c227"
 		},
```

Original report follows.

- - -

Consider adding an audit command which ensures no un-godep'd packages are involved with the current build (thy can be in workspace, but must not be part of the dependnecy graph for package being audited.

```go
package main

import (
	"go/build"
	"flag"
	"fmt"
	"strings"
	"log"
)


var verbose= flag.Bool("v", false, "verbose output")

func init() {
	flag.Parse()
}

func main() {
	ctx := build.Default
	visited := make(map[string]bool)

	pkg, err := ctx.Import(".", ".", build.AllowBinary)
	if err != nil {
		log.Fatalf("error on import: %s", err);
	}			
	explort(ctx, pkg, visited)	
}

func explort(ctx build.Context, pkg *build.Package, visited map[string]bool) {
	for _, packageName := range pkg.Imports {
		if !visited[packageName] {
			visited[packageName] = true
			if *verbose || strings.Contains(packageName, ".") {
				fmt.Printf("%s\n", packageName)
			}
			if ! (packageName == "C") {
				child, err := ctx.Import(packageName, pkg.Dir, build.AllowBinary)
				if err != nil {
					log.Fatalf("error on import: %s", err);
				}					
				explort(ctx, child, visited)
			}
		}
	}
}
```
[...]

Issue #76, opened 2014-04-04
A godep flag or command to check if something needs to be updated


keyneston:
Hey,

It would be useful to have a `godep` flag or command that would let you know if `godep save` would change anything.

Basically return 1 if something needs to be updated or 0 if everything is good. This is useful for writing git pre-commit hooks.

-T

PR #232, opened 2015-06-17
Godep diff implementation

neurogeek:
Implemented the  'godep diff' command which will output a unified diff between existing Godeps and what is now in $GOPATH.
Should close Issue #1 and maybe Issue #76 and Issue #119

### DIFF
--- a/diff.go
+++ b/diff.go
@@ -1,0 +1,80 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+
+	"github.com/pmezard/go-difflib/difflib"
+)
+
+var cmdDiff = &Command{
+	Usage: "diff",
+	Short: "shows the diff between current and previously saved set of dependencies",
+	Long: `
+Shows the difference, in a unified diff format, between the 
+current set of dependencies and those generated on a 
+previous 'go save' execution.
+`,
+	Run: runDiff,
+}
+
+func runDiff(cmd *Command, args []string) {
+	var gold Godeps
+
+	_, err := readOldGodeps(&gold)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	pkgs := []string{"."}
+	dot, err := LoadPackages(pkgs...)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	ver, err := goVersion()
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	gnew := &Godeps{
+		ImportPath: dot[0].ImportPath,
+		GoVersion:  ver,
+	}
+
+	err = gnew.Load(dot)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	diff, err := diffStr(&gold, gnew)
+	if err != nil {
+		log.Fatalln(err)
+	}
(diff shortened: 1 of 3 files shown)


kr:
Rad, looking at this now.

kr:
BTW, maybe I should write this up in a [contributing.md file](https://github.com/blog/1184-contributing-guidelines), but this project has a "rebase & squash" policy. It's fine for a pull request to have multiple commits, but to land it should be rebased on top of master and squashed into a single commit.

diff.go
```diff
@@ -0,0 +1,76 @@
+package main
+
+import (
+	"bytes"
+	"log"
+
+	"github.com/pmezard/go-difflib/difflib"
```
kr: This reminds me, we should probably start using godep on itself soon. Not part of this change, but soon.

diff_test.go
```diff
@@ -0,0 +1,98 @@
...
+	diff, _ = diffStr(&dep1, &dep2)
+
+	if !diffsEqual(strings.Fields(diff), strings.Fields(d2)) {
+		t.Errorf("Expecting diffs to be equal. Obs <%v>. Exp <%v>", diff, d2)
+	}
+}
+
+// diffsEqual asserts that two slices are equivalent.
+func diffsEqual(a, b []string) bool {
+	for i := range a {
```
freeformz: What if b is a different length than b?

Maybe fast path before this?
```go
if len(a) != len(b) { return false }
```

diff.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+var cmdDiff = &Command{
+	Usage: "diff",
+	Short: "shows the diff between current and previously saved set of godeps",
+	Long: `
+Shows the difference, in a unified diff format, between the 
+current set of godeps and those generated on a
```
kr: I suggest s/godeps/dependencies/ because the current set isn't actually saved anywhere yet. As a user I don't conceptualize it as "godeps" until it's written to the file.

diff.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	diff := difflib.UnifiedDiff{
+		A:        difflib.SplitLines(ab.String()),
+		B:        difflib.SplitLines(bb.String()),
+		FromFile: "Godeps",
+		ToFile:   "$GOPATH",
+		Context:  10,
+	}
+	return difflib.GetUnifiedDiffString(diff)
```
kr: Why return this error if it's just going to be ignored above?

diff_test.go
```diff
@@ -0,0 +1,98 @@
...
+		ImportPath: "C",
+		GoVersion:  "go1.2",
+		Deps: []Dependency{
+			{ImportPath: "D101", Comment: "D202"},
+		},
+	}
+)
+
+func TestDiff(t *testing.T) {
+	// Equiv Godeps, should yield and empty diff.
```
kr: s/and/an/

kr:
LGTM after the minor things above.

neurogeek:
@kr and @freeformz Thanks for the comments/review. I've addressed those in my PR.

PR #236, opened 2015-06-29
Don't rely on GOOS and GOARCH

### LINKED ISSUE #235: GoVersion not stripped of GOOS/GOARCH
If godep is compiled and installed with a version of go that is different from the currently running version of go then the GOOS/GOARCH is not stripped from GoVersion,

For instance if godep is compiled and installed with `go version go1.4.2 darwin/amd64` but then later on run when the current go version is `go version go1.4.2 darwin/386`, you end up with a GoVersion of `"GoVersion": "go1.4.2 darwin/386"`.

freeformz:
Split on " " and take the 3rd element instead. Do all go versions follow the
same pattern? Seems like the previous code was basically expecting it already.

This fixes #235 for me.

### DIFF
--- a/dep.go
+++ b/dep.go
@@ -3,12 +3,12 @@ package main
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
-	"runtime"
 	"sort"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


freeformz:
@kr @neurogeek I've already run into this once with a user having previously installed godep via amd64 on darwin and then updating to a newer go version, but using the 386 package instead and then having issues pushing code.  So unless you have any objections wrt the implementation I'm going to merge this at some point.

kr:
SGTM

PR #238, opened 2015-06-30
enable travis-ci

kr:


kr:
PTAL

.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: go
+
+go: 1.4
```
freeformz: Not 1.4.2 ? Or does 1.4 imply the latest from the 1.4.X line?

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+
+script:
+    - git config --global user.email "you@example.com"
+    - git config --global user.name "Your Name"
```
freeformz: Why is this?

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+    - test -z "$(go fmt)"
+    - go vet
+    - go test
+    - go test -race
```
freeformz: Have you considered golint as well?

freeformz:
LGTM modulo questions.

.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: go
+
+go: 1.4
+
+script:
+    - git config --global user.email "you@example.com"
+    - git config --global user.name "Your Name"
+    - test -z "$(go fmt)"
+    - go vet
+    - go test
```
freeformz: I usually add `-v` to my .travis.yml files as well.

freeformz:
@kr I'm going to merge this so things stop failing as Travis seems to default to ruby.

.travis.yml
kr: Godep's unit tests run git to synthesize repositories on disk, and git (at least on travis) requires these to be configured.

.travis.yml
kr: Let's add golint in a separate commit, along with fixing the dozen or so things it complains about.

kr:
Uhh, it would have been nice to be able to address your feedback before we merged this. If we don't want travis complaining while this PR is still open the fix is to turn on this option:

build only if travis.yml is present

I just turned it on.

kr:
Sorry, I guess I had 2 days' worth of chance to address the feedback. :sweat_smile:

My bad.

freeformz:
@kr Also my bad. Was kind of trigger happy.

PR #239, opened 2015-07-01
Inception

### LINKED ISSUE #213: install problem from china
I was in go 1.3.3 . when I run 'go get github.com/tools/godep', it generated a error msg 'unrecognized import pkg golang.org/x/tools/vcs  etc...'.

The domain golang.org is not reachable from mainland in china,because of the chinese gfw :-(

so , would you please copy the golang.org 's vcs package to the github?

freeformz:
Use godep on itself to vendor the few external deps that it has.

This fixes #213 & #214


kr:
LGTM

PR #240, opened 2015-07-02
[WIP] Start playing around with vendor

freeformz:
So far this is mostly a mechanical (sed) replacement of _workspace/src
with vendor.

This is here mostly to play around and doesn't work yet.

### DIFF
--- a/rewrite.go
+++ b/rewrite.go
@@ -109,16 +109,16 @@ func rewriteGoFile(name, qual string, paths []string) error {
 // sep is the signature set of path elements that
 // precede the original path of an imported package
 // in a rewritten import path.
-const sep = "/Godeps/_workspace/src/"
+const sep = "/vendor/"
 
 // unqualify returns the part of importPath after the last
 // occurrence of the signature path elements
-// (Godeps/_workspace/src) that always precede imported
+// (vendor) that always precede imported
 // packages in rewritten import paths.
 //
 // For example,
 //   unqualify(C)                         = C
-//   unqualify(D/Godeps/_workspace/src/C) = C
+//   unqualify(D/vendor/C) = C
 func unqualify(importPath string) string {
 	if i := strings.LastIndex(importPath, sep); i != -1 {
 		importPath = importPath[i+len(sep):]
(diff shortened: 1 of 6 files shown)


kr:
Very exciting! I've also started playing around with this. Maybe there is something here that can help godep https://github.com/kr/vexp.

I suspect the hardest part will be making godep work with both formats, projects saved with older versions of godep as well as the new vendor directory. And possibly also some sort of automatic or explicit migration.

freeformz:
@kr yes. Playing with it I just blew away log-shuttle's Godeps directory and didn't migrate anything.

The other thing that came up is what to do with vendored deps inside of vendored deps. There are several (many?) libraries that can't safely be vendored inside of vendored deps (lib/pq is one example). 

Should they be extracted to a top level? What if your package and a package you vendor both require the same lib. What if they are different versions? Just error out? 

How does all of that affect things like  the cmd/CmdName structure where the root of your package is a library (which should or should not vendor stuff), but the library had deps and the cmd uses the library?

Also not sure how much of this tools like godep/vendor/nut/etc should care about.

kr:
Yeah as far as I know those are still unsolved questions.

I think it's pretty safe to say that the result of vendoring should be a "flat" set of dependencies. That is, no "vendor" directory as a descendant another "vendor" directory. Seems to me like the hardest part is how to choose which version of each dependency.

nathany:
Maybe instead of choose it could act like a git merge. The dependencies conflict and must be manually chosen, or they happen to be identical and automatically do there thing.

Another option might be to take the latest dependency with a warning or requiring a flag, but that would require metadata, yes?

azylman:
Why should the result of vendoring be a "flat" set of dependencies? That seems very confusing - if package A depends on package B and C that both depend on different versions of package D, what version of package D is used?

The spec for vendoring allows for nested vendor folders (https://groups.google.com/d/msg/golang-dev/74zjMON9glU/0e7M7FHipusJ) and you can sidestep that hairy issue if you allow it.

I can see the benefit to "promoting" a vendored dependency upwards if they're the same version, but that seems like an optimization and not necessary.

freeformz:
@azylman While it allows it will cause problems. For instance if multiple things import multiple lib/pq's database/sql will panic because of multiple registrations in the same namespace (making few assumptions here). That is one example of course, but many things assume they'll be initialized once. That is at least one reason, there are probably more.

freeformz:
Closing this in favor of #244 / fe7138c

PR #243, opened 2015-07-08
Verbose output for restore and get commands

zutherb:
Hi,

I had same trouble with godep because my ssh keys was not correct. godep was not able to show my that i could not get access to the repository rather. That's why i added verbose flag to the get and restore commands. I hope this helps someone else.

Regards
Bernd

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
+.idea
 /godep
--- a/get.go
+++ b/get.go
@@ -7,7 +7,7 @@ import (
 )
 
 var cmdGet = &Command{
-	Usage: "get [packages]",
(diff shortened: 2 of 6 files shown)


agileknight:
Any feedback on this?
Would be very useful to have a verbose option as it is easy to get into ssh access and other permission errors with private repos.

kontrafiktion:
please merge

freeformz:
I don't like the output of restore in verbose mode, but it's what get's output by go's tools/vcs stuff. We could clean that up later as I think this is helpful as is, modulo a few changes (which I'm just doing).

freeformz:
Thanks! merged

PR #244, opened 2015-07-10
Respect the GO15VENDOREXPERIMENT env variable

freeformz:
The GO15VENDOREXPERIMENT env variable is how the vendor experiment is
enabled and we should respect that and use it when the user has it set.

@kr ... I think this is something (with maybe a little more work) we should actually incorporate sooner rather than later. Thoughts?

### DIFF
--- a/rewrite.go
+++ b/rewrite.go
@@ -106,16 +106,19 @@ func rewriteGoFile(name, qual string, paths []string) error {
 	return os.Rename(wpath, name)
 }
 
-// sep is the signature set of path elements that
-// precede the original path of an imported package
-// in a rewritten import path.
-var sep = "/Godeps/_workspace/src/"
+// VendorExperiment is the Go 1.5 vendor directory experiment flag, see
+// https://github.com/golang/go/commit/183cc0cd41f06f83cb7a2490a499e3f9101befff
 var VendorExperiment = os.Getenv("GO15VENDOREXPERIMENT") == "1"
 
-func init() {
+// sep is the signature set of path elements that
+// precede the original path of an imported package.
+var sep = defaultSep(VendorExperiment)
+
+func defaultSep(experiment bool) string {
 	if VendorExperiment {
-		sep = "/vendor/"
+		return "/vendor/"
 	}
+	return "/Godeps/_workspace/src/"
 }
 
 // unqualify returns the part of importPath after the last
--- a/rewrite.go
+++ b/rewrite.go
@@ -115,7 +115,7 @@ var VendorExperiment = os.Getenv("GO15VENDOREXPERIMENT") == "1"
 var sep = defaultSep(VendorExperiment)
 
 func defaultSep(experiment bool) string {
-	if VendorExperiment {
+	if experiment {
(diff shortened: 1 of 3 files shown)


kr:
I agree. What else do you think it would need?

rewrite.go
```diff
@@ -109,7 +109,14 @@ func rewriteGoFile(name, qual string, paths []string) error {
 // sep is the signature set of path elements that
 // precede the original path of an imported package
 // in a rewritten import path.
-const sep = "/Godeps/_workspace/src/"
+var sep = "/Godeps/_workspace/src/"
+var VendorExperiment = os.Getenv("GO15VENDOREXPERIMENT") == "1"
+
+func init() {
```
kr: I would suggest the signature `defaultSep(experiment bool) string` for this.

save.go
```diff
@@ -25,7 +25,8 @@ with the exact source control revision of each dependency, and copies
 their source code into a subdirectory.
 
 The list is written to Godeps/Godeps.json, and source code for all
-dependencies is copied into Godeps/_workspace.
+dependencies is copied into either Godeps/_workspace or, if the vendor
+experiment it turned on, vendor.
```
kr: s/it/is/

rewrite.go
freeformz: Hmm. I likened it to flag handling, which is done elsewhere in init functions.

Are you suggesting that init should call `defaultSep(VendorExperiment)` to set sep or that everywhere sep is currently used it should be replaced with `defaultSep(VendorExperiment)` ?

kr: I'm suggesting the first, e.g. `var sep = defaultSep(VendorExperiment)`.

Flags are defined in init functions only because they have to to avoid a cycle. (Try changing it to `var saveR = cmdSave.Flag.Bool(...)` to see what happens.)

freeformz:
@kr Updated with suggestions.

Re: What more. TBH I think this is enough for now. It's usable for me w/o changes to my workflow (modulo https://github.com/golang/go/issues/11659 anyway).

freeformz:
@kr just to be clear I don't think that golang issue should block. Just pointing it out.

kr:
LGTM

freeformz:
@kr I also added docs and fixed an issue, so if you could take another look ++.

kr:
What do you think about putting Godeps.json inside vendor? It would mean more work for other tools that read that file (e.g. the Heroku buildpack) but it would produce less clutter in a project's root directory.

Either way, LGTM

freeformz:
@kr I think I'd rather leave the file where it is. For now at least.

kr:
Looks like this was closed in fe7138c011ae7875d4af21efe8b237f4987d8c4a.

freeformz:
@kr Yeah, sorry I got distracted after the merge with other work.

pierrre:
Does this mean that `godep go build` is equivalent to `go build` if `GO15VENDOREXPERIMENT` is set?

freeformz:
@pierrre Not 100%, but essentially yes.

Issue #24, opened 2013-11-11
dirty working tree


goinggo:
I 'go getted' a package and needed to fix a bug. I don't know if the pull request will be accepted. I want to vendor my version of the code for now.  The tool will not let me.

For now I will need to restore the original, vendor the code, and then make my changes in the vendored version.

Would like to override that protection.

PR #248, opened 2015-07-19
Instruct user on how to resolve "dirty working tree" problem

brandur:
Provides a slightly more helpful error message that continue to present
the problem, but also gives the user a hint on how to resolve it. This
may save some people a trip to Google and back.

@kr I ran into this one and had to land on #24 via Google because it
wasn't too intuitive to me what I should do about it. Would you consider 
this slightly expanded error message to provide a little extra help? 
Thanks!

Helps address #24.

### DIFF
--- a/dep.go
+++ b/dep.go
@@ -120,7 +120,7 @@ func (g *Godeps) Load(pkgs []*Package) error {
 			continue
 		}
 		if vcs.isDirty(pkg.Dir, id) {
-			log.Println("dirty working tree:", pkg.Dir)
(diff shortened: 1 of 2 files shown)


kr:
LGTM

freeformz:
@brandur Thanks! Merged.

qiangli:
Ran into this problem and Google led me here. My "workaround": git stash before godev save.

However, it seems to me too much a restriction. I'd prefer a warning over a hard stop as developers should be able to experiment changes before committing them regardless.

PR #249, opened 2015-07-20
godep version command

freeformz:
Output the version as well as some go runtime information that is
useful for debugging user's issues.

The `version` const would be bumped each time a PR is merged into master
to ensure that we'll be able to tell which version someone got when they
did a `go get github.com/tools/godep`.

This is similar to what github.com/kr/okjson does.

Other suggestions welcome, but I need to introduce the idea of a version
for debugging purposes.

@kr This is what we spoke about @ Gophercon
/cc @neurogeek

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"log"
 	"os"
+	"runtime"
 	"strings"
 	"text/template"
 )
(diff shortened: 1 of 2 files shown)


neurogeek:
I agree with adding version information. How do we enforce remembering to increase the version number, though?

freeformz:
@neurogeek I am not sure, we discussed that @ Gophercon but there wasn't any resolution. Another idea was to use tags and travis to do binary releases to github, but people are still likely to go get godep and we still won't know what version is in use.

Maybe a merge checklist? Even then you can forget to do it.

freeformz:
@neurogeek I'm going to merge this in for now. If we forget to increment the number then at least we know what could possibly be included (last version change and later), where now we don't know.

neurogeek:
No objections from me. Sounds good.

PR #262, opened 2015-08-18
continue try other package instead of exit whole process

lysu:
It's better to continue try other one instead of exit whole process when meet failure duaring `restore` command fetch pkg

In China, our network is under GreatFireWall..

so we need A vpn to get google dep, and B vpn to get others..

it's better do some warn and go on fetch other package..then we can try multi-time to get all deps - -!

I tried to do something like `go get without -u` first see loacal, then decide is need to fetch though network...but it's not easy to implement that so...try this one first~

### DIFF
--- a/dep.go
+++ b/dep.go
@@ -169,8 +169,11 @@ func ReadAndLoadGodeps(path string) (*Godeps, error) {
 		d := &g.Deps[i]
 		d.vcs, err = VCSForImportPath(d.ImportPath)
 		if err != nil {
-			log.Printf("fetch %s failure", d.ImportPath)
-			continue
+			if skipError {
+				log.Printf("fetch %s failure but skipped", d.ImportPath)
+				continue
+			}
+			return nil, err
 		}
 		avs = append(avs, g.Deps[i])
 	}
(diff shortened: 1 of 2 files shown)


lysu:
we add a flag `-skipError` to `restore` command, so it doesn't break the default behave, and skip on error when we supply flags..

freeformz:
AFAICT this doesn't skip at the right place as per your description. It skips errors in parsing the GoDeps file, not on restore.

lysu:
oh~ I meet error first when fetching VCS info of dependency after configuration file parsed but before `restore`,  so I skip unreachable dependency in `ReadAndLoadGodeps` and replace `g.Deps ` with filtered dependency list then `restore` stage will be skip too ..

freeformz:
There are other errors that can occur. Please generalize the message and report the error.

freeformz:
Also needs a rebase.

freeformz:
I am going to close this since there haven't been any updates in ~ 2 weeks or so. Please feel free to re-open with an updated branch if this feature is still important to you. Thanks!

PR #267, opened 2015-09-03
Do not parse Go files in testdata directories.

neurogeek:
Revert ignoring testdata directories and instead, ignore it during the rewrite of Go files and copy the whole directory unmodified.

This commit fixes issue #140

### DIFF
--- a/rewrite.go
+++ b/rewrite.go
@@ -42,6 +42,7 @@ func pkgFiles(pkgs []*Package) []string {
 
 // rewriteTree recursively visits the go files in path, rewriting
 // import statments according to the rules for func qualify.
+// This function ignores the 'testdata' directory.
 func rewriteTree(path, qual string, paths []string) error {
 	w := fs.Walk(path)
 	for w.Step() {
@@ -49,7 +50,11 @@ func rewriteTree(path, qual string, paths []string) error {
 			log.Println("rewrite:", w.Err())
 			continue
 		}
-		if !w.Stat().IsDir() && strings.HasSuffix(w.Path(), ".go") {
+		s := w.Stat()
+		if s.IsDir() && s.Name() == "testdata" {
+			w.SkipDir()
+		}
+		if !s.IsDir() && strings.HasSuffix(w.Path(), ".go") {
 			err := rewriteGoFile(w.Path(), qual, paths)
 			if err != nil {
 				return err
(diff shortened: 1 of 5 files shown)


neurogeek:
@freeformz Would you please take a look at this PR?
Thanks!

rewrite.go
```diff
@@ -49,7 +50,11 @@ func rewriteTree(path, qual string, paths []string) error {
 			log.Println("rewrite:", w.Err())
 			continue
 		}
-		if !w.Stat().IsDir() && strings.HasSuffix(w.Path(), ".go") {
+		s := w.Stat()
+		if s.IsDir() && s.Name() == "testdata" {
+			w.SkipDir()
```
freeformz: Maybe add a `continue` after this?

rewrite.go
```diff
@@ -49,7 +50,12 @@ func rewriteTree(path, qual string, paths []string) error {
 			log.Println("rewrite:", w.Err())
 			continue
 		}
-		if !w.Stat().IsDir() && strings.HasSuffix(w.Path(), ".go") {
+		s := w.Stat()
```
freeformz: For me anyway this makes the intent clearer:

```go
s := w.Stat()
switch s.IsDir() {
  case true:
    if s.Name() == "testdata" {
      w.SkipDir()
      continue
    }
  case false:
    if strings.HasSuffix(w.Path(), ".go") {
      if err := rewriteGoFile(w.Path(), qual, paths); err != nil {
        return err
      }
    }
}
```

Feel free not to change it though.

neurogeek:
Right! Done.

freeformz:
See my comments. Eyeballing the testdata looks good.

neurogeek:
Changed to use <code>switch</code>.

rewrite.go
```diff
@@ -49,10 +50,19 @@ func rewriteTree(path, qual string, paths []string) error {
...
-		if !w.Stat().IsDir() && strings.HasSuffix(w.Path(), ".go") {
-			err := rewriteGoFile(w.Path(), qual, paths)
-			if err != nil {
-				return err
+		s := w.Stat()
+		switch s.IsDir() {
+		case true:
+			if s.Name() == "testdata" {
+				w.SkipDir()
+				continue
```
freeformz: heh. in this version the continue is superfluous AFAIK

neurogeek:
Sigh, indeed. 
I forgot the non-fall-through behavior in Go's <code>switch</code> statement.

freeformz:
@neurogeek I rebased and merged this as aeda8bab6aa7e64e94a83b40e29858daeb85ee87

neurogeek:
Awesome! Thanks @freeformz

PR #275, opened 2015-09-14
Consolidate godep file handling

freeformz:
Mostly consolidates godep file handling to a single location, instead of each command doing it it's own, slightly different way.

There are some other small tweaks in here related to naming *Godeps.Load -> *Godeps.fill

/cc @neurogeek @kr

### DIFF
--- a/dep.go
+++ b/dep.go
@@ -1,26 +1,11 @@
 package main
 
 import (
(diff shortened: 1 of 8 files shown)


kr:
Sure, LGTM

PR #276, opened 2015-09-15
Don't save test dependencies by default

freeformz:
Switch to not saving test files by default.

Add `-t` for those wishing to analyze and save test dependencies.

This is important for the usability of the vendor/ experiment where
`./...` also matches everything in `vendor/`.

You should test your dependencies, but it's not necessarily appropriate
to test them every time you want to test all of the code in your repository.

If you still want to save test files add `-t` to your save command(s).

### DIFF
--- a/save_test.go
+++ b/save_test.go
@@ -65,6 +65,7 @@ func TestSave(t *testing.T) {
 		cwd      string
 		args     []string
 		flagR    bool
+		flagT    bool
 		start    []*node
 		altstart []*node
 		want     []*node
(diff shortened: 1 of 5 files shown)


freeformz:
@neurogeek @kr this is on top of #275.

amerine:
@freeformz I like it!

save_test.go
```diff
@@ -1008,6 +1009,72 @@ func TestSave(t *testing.T) {
...
+				{"C/Godeps/_workspace/src/D/main.go", pkg("D"), nil},
+			},
+			wdep: Godeps{
+				ImportPath: "C",
+				Deps: []Dependency{
+					{ImportPath: "D", Comment: "D1"},
+				},
+			},
+		},
+		{ // don't copy _test.go files
```
kr: s/don't/do/
?

freeformz: Oops fixed.

kr:
LGTM

eparis:
So this merged.  But It seems to be slightly overzealous.  I have dependencies in MY _test.go files which are not in my .go files.  `godep save` no longer pulls in those dependencies. So `godep go test` doesn't work any more.  I love not pulling OTHER people's _test.go, but I still want it to pull in whatever I need in my _test.go...

freeformz:
@eparis Doh! I fixed this in v10. Sorry about that.

PR #281, opened 2015-09-22
Add changelog

buddhamagnet:
Wondered if it would make sense to add a changelog so people can see what's included in each release? Happy to help maintain this is it's a help?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,3 +10,17 @@
 
 * Analyse vendored package test dependencies.
 * Update documentation.
+
+# v9 17/09/2015
+
+* Don't save test dependencies by default.
+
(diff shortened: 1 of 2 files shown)


freeformz:
Sure, care to update this and add a 'v' in front of the version? If you can back track this a little as well that would be good, but not 100% necessary.

buddhamagnet:
Have done my best, not easy as the version bumps are not always contained in a single commit.

buddhamagnet:
OK think that's pretty much it, let me know otherwise!

freeformz:
@buddhamagnet looks great! I squashed them all down to a single commit and merged them. Thanks!

PR #283, opened 2015-09-22
Split command struct out into own file

buddhamagnet:
Split command type and methods into separate file for clarity.

### DIFF
--- a/command.go
+++ b/command.go
@@ -1,0 +1,60 @@
+package main
+
+import (
+	"flag"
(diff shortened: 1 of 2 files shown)


freeformz:
I think Command belongs in main as it's mostly involved with the setup and execution of the various commands anyway. Good place to keep all the code, at least for now.

buddhamagnet:
Understood, thanks.

PR #284, opened 2015-09-22
Extract errors into central file

buddhamagnet:
Extracted some duplicate error values into a separate file for easier re-use. Hope all this activity helps, I love godep and love contributing!


freeformz:
Thanks!

PR #285, opened 2015-09-29
restore: download all in 1 pass, then restore

### LINKED ISSUE #286: godep restore results in master instead of revision specified in Godeps.json for certain dependencies
We've been debugging https://github.com/kubernetes/kubernetes/issues/14677 and have discovered that if you have a set of dependencies like this:

google.golang.org/api/compute/v1, revision 0c2979aeaa5b573e60d3ddffe5ce8dca8df309bd
golang.org/x/net, revision cbcac7bb8415db9b6cb4d1ebab1dc9afbd688b97

What we're seeing is that because `godep restore` performs `go get -d` and then e.g. `git checkout <rev>` for each dependency, it's possible to end up in a state where 1 dependency has the correct revision, while the other is on master. For example:

1. `go get -d golang.org/x/net` (gets HEAD aka master)
1. `git checkout cbcac7b`
1. `go get -d google.golang.org/api/compute/v1` (gets HEAD aka master)

At this point, when retrieving the head of google.golang.org/api/compute/v1, it includes a dependency on `golang.org/x/net/context/ctxhttp`. This particular directory is not part of `golang.org/x/net` in revision cbcac7b, As a result, `go get` fixes the detached head for golang.org/x/net and then invokes `git pull --ff-only` to (hopefully) pull in the missing directory. This succeeds, and `godep restore` then finally does `git checkout 0c2979a`. What you're left with is this:

google.golang.org/api/compute/v1, revision 0c2979aeaa5b573e60d3ddffe5ce8dca8df309bd
golang.org/x/net, revision master

I have created #285 as an attempt to fix this. It performs restore in 2 phases: 1) download all the packages 2) checkout the appropriate revisions.

Please let me know your thoughts on this and if there's a better way to fix it.

ncdc:
Modify restore so that it downloads all the dependencies in one pass,
and then runs restore for each dependency in a second pass. We have seen
an issue where, depending on the order of the `go get` downloads,
sometimes one or more dependencies end up with the master branch as
HEAD, instead of the revision specified in Godeps.json. It appears that
`go get` resets the HEAD to master, even if the dependency has already
been downloaded and is pointing to a different ref.

Fixes #286

### DIFF
--- a/restore.go
+++ b/restore.go
@@ -27,10 +27,19 @@ func runRestore(cmd *Command, args []string) {
 	}
 	hadError := false
 	for _, dep := range g.Deps {
-		err := restore(dep)
+		err := download(dep)
 		if err != nil {
 			log.Println("restore:", err)
 			hadError = true
+		}
+	}
+	if !hadError {
+		for _, dep := range g.Deps {
+			err := restore(dep)
+			if err != nil {
+				log.Println("restore:", err)
+				hadError = true
+			}
 		}
 	}
 	if hadError {
@@ -38,9 +47,8 @@ func runRestore(cmd *Command, args []string) {
 	}
 }
 
-// restore downloads the given dependency and checks out
-// the given revision.
-func restore(dep Dependency) error {
+// download downloads the given dependency.
+func download(dep Dependency) error {
 	// make sure pkg exists somewhere in GOPATH
 
 	args := []string{"get", "-d"}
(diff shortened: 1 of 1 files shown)


ncdc:
I'm not sure if we're doing something wrong with our use of Godep, or if there's a bug in `go get`, but we're running into https://github.com/kubernetes/kubernetes/issues/14677. Please let me know if there's a better way to go about this - thanks!

karlkfi:
I don't remember this being a problem before. Was `go get -d` changed at some point to re-up dependencies? I could have sworn it used to leave them alone when they were on a detached head.

I think this is probably the correct approach for now. The primary revision should override the transitive one. Tho it's probably a good idea to warn when this is the case, either when restoring or saving or both.

ncdc:
If we want to fix this without changing the godep code, if we modify the
script to call godep restore twice, that should take care of the problem as
well.

karlkfi:
Based on the analysis in #286 this is cased not just by having a transitive dependency, but having a transitive dependency on a new package that is a sub-package of a direct dependency with a revision where the sub-package did not exist.

freeformz:
@ncdc sorry for the delay, I was on vacation for a little over a week.

AFAICT there would also be an issue with resolving dependencies as well and recording the correct one.
This would be required anyway though.

So LGTM, merged!

PR #290, opened 2015-10-08
Do a release to github

### LINKED ISSUE #110: A way to install godep securely, hashable/pinning
I like godep because it allows me to install dependencies securely, specifying the hash of each dependency. But I need a way to bootstrap that. go-lang is installed in my system using a package manager like apt, but to install godep from an official source I have to use "go get github.com/tools/godep", which is not very secure. Couldn't you do releases and package them, so that we can download and unpack them and install them in GOPATH for bootstraping?

I'd like to be able something like:
$ mkdir gopath; cd gopath; wget http://path/to/godep-rev.tar.gz
$ [ "$(sha512sum godep-rev.tar.gz)" == "whateverhash" ] && tar zxf godep-rev.tar.gz || echo "Invalid tarball"
$ export GOPATH="$(pwd)"
$ then use godep

freeformz:
For those that want officially created binaries instead of requiring `go
get` as the only install method.

I would remove the `all_branches` flag if merged so that it only works
off master, but I need it now for testing.

Fixes #110 (or at least part of it, I don't intent to cover the hash / security parts).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,10 @@ script:
 before_install:
 - go get golang.org/x/tools/cmd/vet golang.org/x/tools/cmd/goimports
 before_deploy:
+- export TARGETS="linux darwin windows"
 - go get github.com/mitchellh/gox
-- gox -build-toolchain
-- gox -os "linux darwin windows"
+- gox -os "$TARGETS" -build-toolchain
+- gox -os "$TARGETS"
 deploy:
   skip_cleanup: true
   provider: releases
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,5 @@ deploy:
     - godep_windows_386.exe
     - godep_windows_amd64.exe
   on:
-    all_branches: true
-    #tags: true
+    tags: true
     repo: tools/godep
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ script:
 before_install:
 - go get golang.org/x/tools/cmd/vet golang.org/x/tools/cmd/goimports github.com/mitchellh/gox
 before_deploy:
+- gox -build-toolchain
 - gox -os "linux darwin windows"
 deploy:
   skip_cleanup: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,5 +28,5 @@ deploy:
     - godep_windows_amd64.exe
   on:
     all_branches: true
-    tags: true
-    repo: tools/godep1p
+    #tags: true
+    repo: tools/godep
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,9 @@ script:
 - go test -v -race
 - test -z "$(goimports -l .)"
 before_install:
-- go get golang.org/x/tools/cmd/vet golang.org/x/tools/cmd/goimports github.com/mitchellh/gox
+- go get golang.org/x/tools/cmd/vet golang.org/x/tools/cmd/goimports
 before_deploy:
+- go get github.com/mitchellh/gox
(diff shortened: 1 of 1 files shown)


freeformz:
@kr @neurogeek thoughts?

kr:
This feature sounds reasonable in principle, just for the sake of convenience. (I would be careful about doing things for the sake of security without thinking more deeply about it from top to bottom. I suspect that godep isn't in a position to provide any meaningful security guarantees without some serious redesign. A feature that is nominally "for security" but where the threat model is half-baked or nonexistent will be misleading, and probably worse than just making no promises in the first place.)

I don't have much opinion about the mechanism to achieve it. If this works reliably, cool.

.travis.yml
```diff
@@ -1,19 +1,35 @@
...
+- export OS_TARGETS="linux darwin windows"
+- export ARCH_TARGETS="386 amd64"
+- go get github.com/mitchellh/gox
+- gox -os "$OS_TARGETS" -arch="$ARCH_TARGETS" -build-toolchain
+- gox -os "$OS_TARGETS" -arch="$ARCH_TARGETS"
+deploy:
+  skip_cleanup: true
+  provider: releases
+  api_key:
+    secure: Q1JP8LziaXMTxFmNXiyC1YhS9e4M4WnI6UDjRTMf6mm1LZeJyUFOCCtXnifL7RyCIR1hpjp6s8M1aWE+NpuweF96IZI3Uk4ASx5C8FePC4qvhsCdtJ2sLD2GTIrp9b0MS9/+ao20AIbpVDSaLaF9IjqXpMxMyM0P8P5coRTkwItlGxmQbVJW3YuiYcPa8UojwM4EyafO2CIoUKapW8lwb9KcimBJV8PfF/XZjPVhMkn2ABhh5Hqbn2zBJtvPYMMzi0CnY50JQF5LwN3vGTMpTsRP+lOLCNbOWfkl+2hgG7VpKrtx+cX62knOodpF457sIJ31KUzmeLUVBejTGb1zuVeTojuyi8Huo8YBIBCcN+p3Dqd+n2ZK45mIrheGiEJIkf/vI4MI6A01Nu/o+xU0IPsVfAL/xU5j5nntEGfFWVoclPrl9qcfqf74xdRcARzcCJVmdc8iw49DBDHJfnPa3zxzVz//00+Rz6mZXmhk+Npk/HLLNW59vmJIjP+8XOtPor7dST9HrS1a9AcnmIjNuw9yfbwK5769SDVxCKgqNwXW/Dy5F39aIH5AL4I4y9hCEeeT8ctvSJHGOyiB9MWU5jnt5tluPtz5opG51tFXnIYP/XaWpTfO+eJ6x55pbwT+n3LfRS5l1POM+jGAFF1MFWwc14RY7qynEIEzm4Wb/UE=
```
kr: Does committing this secret to git mean that anyone on earth can publish a release binary? That seems like a problem.

freeformz: They are encrypted using a private key associated with this repo so I believe they won't decrypt properly for another repo / fork / etc (as per Travis docs).

freeformz:
@kr Re: Security. My goal is to just automate the creation of releases when tags are pushed. I don't intend to tackle the other parts of #110 really for all of the reasons you mention. IMO for "normal" use cases trusting Travis + GitHub and their certs is probably fine for the 90% of people wanting to curl down a release. I understand that isn't "secure" though in anyway.

neurogeek:
Just to pitch in.
I like the idea of official releases, for convenience that is. Security would be another matter.

freeformz:
+1, merging because this isn't meant to address the security part.

Issue #86, opened 2014-05-19
improve version mismatch error message


kr:
From #85, we have a badly-written error message that occurs when godep save can't carry forward existing versions. It should be rewritten to be more helpful.

PR #291, opened 2015-10-11
save: Improve error message when trying to save a conflicting revision.

tsuna:
When a sub-path is saved, godep requires that all subpaths originating
from the same repo be pinned at the same revision.  When one attempts to
save a new subpath at a different revision, godep produces a confusing
error message, which this commits improves.

In my case I had golang.org/x/net already saved and as a result of
pulling new gRPC code, `godep save ./...` was trying to save the new
path golang.org/x/net/http2, and failing in a puzzling way:

> $ godep save ./...
> godep: golang.org/x/net/http2: revision is b846920a172af75fe52c1400ae6094307be83b8a, want 7654728e381988afd88e58cabfd6363a5ea91810

Now:

> $ godep save ./...
> godep: cannot save golang.org/x/net/http2 at revision 7654728e381988afd88e58cabfd6363a5ea91810: already have golang.org/x/net at revision b846920a172af75fe52c1400ae6094307be83b8a.
> Update golang.org/x/net first.

### DIFF
--- a/save.go
+++ b/save.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"log"
@@ -186,12 +187,14 @@ func save(pkgs []string) error {
 
 type revError struct {
 	ImportPath string
+	HavePath   string
 	HaveRev    string
 	WantRev    string
 }
 
 func (v *revError) Error() string {
-	return v.ImportPath + ": revision is " + v.HaveRev + ", want " + v.WantRev
+	return fmt.Sprintf("cannot save %s at revision %s: already have %s at revision %s.\n"+
+		"Run `godep update %s' first.", v.ImportPath, v.WantRev, v.HavePath, v.HaveRev, v.HavePath)
 }
 
 // carryVersions copies Rev and Comment from a to b for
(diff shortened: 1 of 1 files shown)


kr:
:+1: this is a big improvement.

One could nit-pick the wording or formatting, but this is already much better than what it currently does.

tsuna:
I'm open to any concrete suggestions regarding the wording/formatting.

freeformz:
:+1: 

@tsuna how about telling the user "how" to update first?

freeformz:
May also want to indicate the relationship between the two packages. "packages belong to same repo" or some such?

save.go
```diff
@@ -186,12 +187,14 @@ func save(pkgs []string) error {
...
 	ImportPath string
+	HavePath   string
 	HaveRev    string
 	WantRev    string
 }
 
 func (v *revError) Error() string {
-	return v.ImportPath + ": revision is " + v.HaveRev + ", want " + v.WantRev
+	return fmt.Sprintf("cannot save %s at revision %s: already have %s at revision %s.\n"+
+		"Update %s first.", v.ImportPath, v.WantRev, v.HavePath, v.HaveRev, v.HavePath)
```
freeformz: perhaps...
```go
"`go update %s` first.", ....
```` 
?

freeformz:
This should also fix #86

tsuna:
I amended my change as per the feedback above.

tsuna:
Oops didn't mean to close the PR, sorry, fat finger / wrong button.

freeformz:
@tsuna thanks. merged.

PR #300, opened 2015-10-22
-no-skip and -skip options for godep save

krhubert:
When working with internal projects there is no need to keep all vendors for internal project. The skip option is good way to remove them from Godeps.


freeformz:
This would make the code unbuildable w/o those packages, so I can't accept this.

Issue #245, opened 2015-07-16
godep unintentionally stripping license files from vendored libraries when target is a subdirectory


cactus:
It appears godep strips license files and other data from dependencies when godep save is used.
I have a [project](https://github.com/cactus/go-statsd-client) which uses a nested subdirectory as the main entry point for most folks. When godep save is used by consumers of the library, this is the resulting file hierarchy.

```
Godeps/_workspace/src/github.com/cactus/go-statsd-client
Godeps/_workspace/src/github.com/cactus/go-statsd-client/statsd
Godeps/_workspace/src/github.com/cactus/go-statsd-client/statsd/buffered.go
Godeps/_workspace/src/github.com/cactus/go-statsd-client/statsd/buffered_bench_test.go
Godeps/_workspace/src/github.com/cactus/go-statsd-client/statsd/buffered_test.go
Godeps/_workspace/src/github.com/cactus/go-statsd-client/statsd/doc.go
Godeps/_workspace/src/github.com/cactus/go-statsd-client/statsd/main.go
Godeps/_workspace/src/github.com/cactus/go-statsd-client/statsd/main_bench_test.go
Godeps/_workspace/src/github.com/cactus/go-statsd-client/statsd/main_test.go
Godeps/_workspace/src/github.com/cactus/go-statsd-client/statsd/noop.go
Godeps/_workspace/src/github.com/cactus/go-statsd-client/statsd/test-client
Godeps/_workspace/src/github.com/cactus/go-statsd-client/statsd/test-client/main.go
```

Note the lack of a LICENSE file which is present in the original repository. This means projects using the repo in conjunction with godep would be in non-compliance with the license, wholly unintentionally.

There are several other projects which use similar hierarchy layouts. [Here](https://github.com/cloudflare/cfssl) is one such example.

freeformz:
Hmm. Godep uses go's own libraries/tools to handle package/file selection, which basically ignores everything that isn't a *.go file. This is unlikely to be a "trivial" fix. :-(

kr:
It should be copying all files, not just `.go` source files. See copySrc in save.go.

kr:
Aha, it seems the license file is in a parent directory of all the go packages. Since godep copies only packages, it misses the parent directory. Yes, this would be difficult to fix.

[...]

PR #301, opened 2015-10-22
Add functions to test if a file likely to be legal notice tools/godep#245

client9:
Issue #245 

Currently un-used functions to test if a file is a likely to be software license or legal file.

I'm happy to take a stab a solve the copying problem but I'm not that familiar with the godep code and will need a pointer to the right area.

regards,

n

### DIFF
--- a/save.go
+++ b/save.go
@@ -265,7 +265,19 @@ func removeSrc(srcdir string, deps []Dependency) error {
 	return nil
 }
 
+// parents returns list of parent directories given a path
+// e.g. parent("a/b/c") -> []string{"a/b", "a"}
+func parents(path string) []string {
+	out := []string{}
+	for path = filepath.Dir(path); path != "."; path = filepath.Dir(path) {
+		out = append(out, path)
+	}
+	return out
+}
+
 func copySrc(dir string, deps []Dependency) error {
+	// mapping to see if we visited a parent directory already
+	visited := make(map[string]bool)
 	ok := true
 	for _, dep := range deps {
 		srcdir := filepath.Join(dep.ws, "src")
@@ -279,6 +291,8 @@ func copySrc(dir string, deps []Dependency) error {
 			log.Println(err)
 			ok = false
 		}
+
+		// copy actual dependency
 		vf := dep.vcs.listFiles(dep.dir)
 		w := fs.Walk(dep.dir)
 		for w.Step() {
@@ -288,10 +302,35 @@ func copySrc(dir string, deps []Dependency) error {
 				ok = false
 			}
 		}
+
+		// check parent directorys for legal files and copy
+		for _, parent := range parents(rel) {
+			depdir := filepath.Join(srcdir, parent)
+
+			// prevent copying twice
+			if visited[depdir] {
+				continue
+			}
+			visited[depdir] = true
+
+			vf := dep.vcs.listFiles(depdir)
+			w := fs.Walk(depdir)
+			for w.Step() {
+				if IsLegalFile(w.Stat().Name()) {
+					err = copyPkgFile(vf, dir, srcdir, w)
+					if err != nil {
+						log.Println(err)
+						ok = false
+					}
+				}
+			}
+		}
 	}
+
 	if !ok {
 		return errorCopyingSourceCode
 	}
+
 	return nil
 }
 
--- a/save.go
+++ b/save.go
@@ -265,16 +265,6 @@ func removeSrc(srcdir string, deps []Dependency) error {
 	return nil
 }
 
-// parents returns list of parent directories given a path
-// e.g. parent("a/b/c") -> []string{"a/b", "a"}
-func parents(path string) []string {
-	out := []string{}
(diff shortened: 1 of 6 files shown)


freeformz:
Looks like a good start.
Any reason why we shouldn't include this as a library though?

Wrt using it ...
godep (generally speaking) uses absolute paths to files it finds in the repo. [copySrc](https://github.com/tools/godep/blob/32d49a17abe34aec8a58f6b4d049efd846206dfe/save.go#L268-L296) uses [vcs.listFiles()](https://github.com/tools/godep/blob/32d49a17abe34aec8a58f6b4d049efd846206dfe/vcs.go#L124-L144)  to get that list and [copyPkgFile](https://github.com/tools/godep/blob/32d49a17abe34aec8a58f6b4d049efd846206dfe/save.go#L298-L328) to conditionally copy the files into the vendor directory.

So IMO `IsLicenseFile` and `IsLegalFile` should take into consideration that it may get a full (or partial) path to a filename, not just the basename

client9:
Oh you can include it as a library.  Maybe thats better since godep isn't really structured as a library (thats too bad since the VCS stuff might be nice to reuse else where).  I was under the (wrong) impression you didn't use external dependencies in godep.  I will work on cleaning up my library.

I just added some code to copySrc that seems to do the right thing.  But Im new to this code base.  Suggestions most welcome.

I'll make a fake app and try it out with cactus's statsd library.

re: So IMO IsLicenseFile and IsLegalFile should take into consideration that it may get a full (or partial) path to a filename, not just the base name


thanks for tips.  I'll take a look at other calls to copySrc

onward

freeformz:
@client9 vcs.listFiles() already produces a list of all possible filenames (with absolute paths) that would be considered for copying. Each of those is passed to copyPkgFile, where filtering takes place (arguably the filtering should be in a separate function) and if not filtered out the file is then copied. Please take a look at those functions as you can probably just hook into copyPkgFile.

client9:
ahh great.  Will look!

client9:
Hmm, my understanding of the code is that if it's a normal dependency directory, then
*everything* gets copied unless it ends in `_test.go` (sometimes).

From reading the ticket, for the parent directories, we want *nothing* copied unless it's legal file.

Options:
* pass in another flag to `copyPkgFile` and sort this out in this function.  Thats kinda messy since most of logic doesnt really apply.
* write a copyLegalFile function which is a simplified version of  `copyPkgFile`  It will copy a few fines, but mostly be smaller and simpler.

any preference or did I miss something here?

freeformz:
@client9 After reading your comment I realized that [dep.dir](https://github.com/tools/godep/blob/32d49a17abe34aec8a58f6b4d049efd846206dfe/save.go#L282) could be a sub package as well, which I didn't consider last night, so you'll only be [walking](https://github.com/tools/godep/blob/32d49a17abe34aec8a58f6b4d049efd846206dfe/save.go#L284-L290) through the files in that sub package.

Another loop, using the contents of [vf](https://github.com/tools/godep/blob/32d49a17abe34aec8a58f6b4d049efd846206dfe/save.go#L282), which contains all tracked files in the repo (as absolute paths) is possibly the way to go?

client9:
thanks for your review.

re: "After reading your comment I realized that dep.dir could be a sub package as well, which I didn't consider last night, so you'll only be walking through the files in that sub package."

True, but legal files in a sub package are copied already.  Its really just the parents that is a problem.  So the current pull solves the ticket issue (although we can clean up the patch a bit).  Or are we solving different problems ?

client9:
Ahh I put my comments in the original ticket.  Sorry for the confusion.

https://github.com/tools/godep/issues/245#issuecomment-150476193

save.go
```diff
@@ -288,10 +302,35 @@ func copySrc(dir string, deps []Dependency) error {
 				ok = false
 			}
 		}
+
+		// check parent directorys for legal files and copy
+		for _, parent := range parents(rel) {
```
freeformz: There's no need to try to find the parent, vf already contains all files for the repo that contains the package. I realized that I didn't update the listFIles() doc comment. So I did: 2115b735b60387f29297a05fcfed85c77e40530f. Was that a source of confusion?

client9:
oh @freeformz.. mostly me posting in issue and in pull request, and learning how you guys like doing stuff, and a new code base.  Nothing you did.  I'll take a look more closely...

client9:
I played around with vcs and friends and finally understand the hint you are giving....  and realized I was "solving" a more general problem of a license is any parent directory.   We just assume if the sub-package is missing a license, the license (if any) better be at the top of the package.  No need to walk up.

freeformz:
@client9 Cool, let me know when I should re-review the PR.

client9:
hi @freeformz   latest should have a simplified version, although I think it can be simplified further.  Let me know what you think.

freeformz:
@client9 Thanks, merged after a rebase/squash/version/changelog addition.

freeformz:
dcbe29c7a18887497b4857e64665d25be43f7c7e

Issue #303, opened 2015-10-26
Godep not handling nested packages cleanly


Alex-Addy:
When having a package inside a git repo alongside another package `godep save` will fail when the git repo state is dirty.

The situation is just about

```
.
|--dirty
|    |----Godeps
|    |        \...
|    \ main.go
|--help
    \ help.go
```

With help being in its own package and main using that package. The problem comes down to being unable to have a dirty main and run the `godep save` command, as the git state is global to a repo and a dirty file anywhere will cause `help.go` to be considered dirty. This makes it very difficult to use go's packages to split up functionality 

Also godep will put help into the Godeps directory. Causing it to be duplicated inside the same repo.

I have made a repo to demonstrate this at https://github.com/Alex-Addy/godep_nested_packages.git
To reproduce clone the repo, modify main.go (ex `echo "//" >> main.go`), then run `godep save` inside dirty.

Are there any suggestions on how to get around this? Or maybe an explanation on why it was considered necessary to do this check?

A possible solution would be to allow for skipping the `vcs.IsDirty` check in general or on those packages in the same repo via a command line flag.

freeformz:
What is the vcs root in the above example? I assume it's the parent directory of `dirty` and `help` as in the example?

rasky:
This should be fixed by my pull #302

freeformz:
@rasky I am trying to understand what's going on here and the motivators. Godep is designed to be run in the top package of a repo. There is nothing ATM in your PR BTW.

rasky:
What you mean "there's nothing"? It's a one character patch. It's small, but it's effective :) I think my pull description tries to explain the issue, but I can try again. There are many git repos which are COLLECTIONS of packages, not just single packages. This is especially visibile in /x/ repos. If you import `golang.com/x/crypto/ssh` and run go get, you will get a full checkout of `golang.com/x/crypto` which is a single git repository including dozens of different packages (see: https://github.com/golang/crypto). Now let's say that for another project you locally modify a file within `golang.com/x/openpgp`, in your GOPATH. If you later run `godep save` on the first project (the one using `golang.com/x/crypto/ssh`), you get a dirty error by godep, because the modification in your copy of `golang.com/x/openpgp` is wrongly seen as a modification affecting your project; but your project uses `golang.com/x/crypto/ssh`, not `golang.com/x/crypto/openpgp`, so there is not need to abort the `godep save` operation.

My one-character PR fixes this.

freeformz:
[...]

PR #305, opened 2015-10-29
Adds a -i flag to override the VCS dirty state check for development purposes.

jdmcfad:
Our company has multiple packages in a single git repo, and we hit a roadblock with godep when we pulled some code out of one pakage into an imported package in an adjacent directory.  The zealous `isDirty` check prevents us from running `godep save` when we are working with changes that affect multiple packages.

Note that our `godep save` issue is similar to rasky's issue from his pull request (rasky/godep@73ef0c2cddab0bb26773e5b387b2b446732dd552) but I believe @rasky is dealing with unrelated packages, whereas ours _are_ related.

### DIFF
--- a/save.go
+++ b/save.go
@@ -17,7 +17,7 @@ import (
 )
 
 var cmdSave = &Command{
-	Usage: "save [-r] [-v] [-t] [packages]",
+	Usage: "save [-r] [-v] [-t] [-i] [packages]",
 	Short: "list and copy dependencies into Godeps",
 	Long: `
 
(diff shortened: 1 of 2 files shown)


Alex-Addy:
This would allow for differentiating between development and release builds in the build process. closes #303.

freeformz:
When you say 'adjacent' directory, what do you mean. Can you give me an example package directory structure?

jdmcfad:
See the situation @Alex-Addy outlined in #303 for example.  Does that make sense?

jdmcfad:
This is OBE by `GO15VENDOREXPERIMENT` adopted by Go 1.5/1.6, we haven't looked at this since December.  Thanks for looking at this.

PR #307, opened 2015-11-03
Fix error when having conflicting revisions

freeformz:
These cases seem to simplify down to the same error.

Also fixes the ordering during construction of the error, which was
wrong.


shurcooL:
> Also fixes the ordering during construction of the error, which was
wrong.

That parts looks good to me, but I am not familiar with the rest to review it.

PR #311, opened 2015-11-05
Do not use --debug to find hg identify revision

ontsysmattholcomb:
When an hg extension uses a post-identify hook it also outputs a debug statement notifying that the hook has executed.  godep uses 'identify --id --debug' to get the current full revision identifier.  The --debug causes the post-identify hook debug statement to be output which godep sees as being a 'dirty working tree' and fails


ontsysmattholcomb:
Thanks!

freeformz:
No, thank you!

PR #314, opened 2015-11-06
run command once during restore -v

jmhodges:
Running `godep restore -v` would always error out because the CombinedOutput command did a Run under the hood, and the second Run at the end of restore would error out.
    
Along the way, I added some more context (and a couple of newlines) to the restore error prints to figure out what was going on.

### DIFF
--- a/restore.go
+++ b/restore.go
@@ -29,7 +29,7 @@ func runRestore(cmd *Command, args []string) {
 	for _, dep := range g.Deps {
 		err := download(dep)
 		if err != nil {
-			log.Println("restore:", err)
+			log.Println("restore, during download dep:", err)
 			hadError = true
 		}
 	}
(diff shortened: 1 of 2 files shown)


freeformz:
how about this instead:?

```go
func runIn(dir, name string, args ...string) error {
	c := exec.Command(name, args...)
	c.Dir = dir
	output, err := c.CombinedOutput()

	if verbose {
		fmt.Printf("execute: %+v\n", c)
		fmt.Printf(" output: %s\n", output)
	}

	return err
}
```

Seems simpler and maintains intent AFAICT.

jmhodges:
Yeah, I should have pulled back and taken a better look. Thanks!

jmhodges:
Okay, this is ready for review again!

freeformz:
@jmhodges Thanks, merged as 605fb21360e22cca107bdc4d0c972cbf14353977 (I assume due to a rebase).

PR #318, opened 2015-11-09
add -m flag to support only major version output

jtarchie:
Related to this [issue](https://github.com/heroku/heroku-buildpack-go/issues/100).

[#107206100]
https://www.pivotaltracker.com/story/show/107206100

Signed-off-by: JT Archie <jtarchie@gmail.com>

### DIFF
--- a/save.go
+++ b/save.go
@@ -17,7 +17,7 @@ import (
 )
 
 var cmdSave = &Command{
-	Usage: "save [-r] [-v] [-t] [packages]",
+	Usage: "save [-r] [-v] [-t] [-m] [packages]",
 	Short: "list and copy dependencies into Godeps",
 	Long: `
 
@@ -101,6 +101,8 @@ func save(pkgs []string) error {
 	if err != nil {
 		return err
 	}
+
+	ver = strings.Join(strings.Split(ver, ".")[:2], ".")
 
 	gold, err := loadDefaultGodepsFile()
 	if err != nil {
--- a/save_test.go
+++ b/save_test.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"reflect"
+	"runtime"
 	"strings"
 	"testing"
 	"text/template"
@@ -61,6 +62,7 @@ func godeps(importpath string, keyval ...string) *Godeps {
 }
 
 func TestSave(t *testing.T) {
+	var majorGoVersion = strings.Join(strings.Split(runtime.Version(), ".")[:2], ".")
 	var cases = []struct {
 		cwd      string
 		args     []string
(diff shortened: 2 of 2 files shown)


save.go
```diff
@@ -52,6 +52,8 @@ vendor experiment.
 
 If -v is given, verbose output is enabled.
 
+If -m is given, only save out to the minor go version to Godeps.json.
```
kr: s/minor/major/

For example, Go 1.5 is a major release, and 1.5.1 is a minor revision.
See https://golang.org/doc/devel/release.html.

kr:
Is there a way to do this without a flag? Optional behavior makes for [poor UX and harder-to-maintain software](http://ometer.com/free-software-ui.html).

jtarchie:
@kr thanks for the feedback with minor and major. We'll gladly change the copy.

As a follow up to the `-m` flag, I'd like to make a clarification. We wanted to maintain backwards compatibility with the current `godep` workflow. The flag was introduced to make it opt-in.

Would the desired work flow then, not be a flag, but a new command (e.g. `godep major-version-only`) rather than an extension to `godep save`? That would make it more explicit through the command line.

Happy to chat!

kr:
I can sympathize with the instinctual desire to maintain backwards compatibility. It's worth asking whether that's necessary. I don't know the answer, but if it's acceptable to change the meaning of godep save, then I think that would be preferable.

I'm only aware of one thing that consumes this field, and it's buildpacks (both heroku and cloudfoundry) that need to install the go compiler. If that's the extent of it, then I think it would be fine to simply make the change unilaterally. There might be additional ways this field is used, but I sort of doubt it.

@freeformz do you have any insight here?

jtarchie:
I think we are alright with that. Let me bring it up internally.

freeformz:
I think I'm alright with is as well (which is a change from my initial reaction).  Asking around a little internally as well.

I would like to land a change to heroku/heroku-buildpack-go to select the latest minor version of the specified major version before merging a PR that implements @kr's suggestion.

jtarchie:
@kr and @freeformz, we changed the default `godep save` behaviour to always emit the major version only.

As for the buildpack support, @freeformz why were you thinking this has to come in after the buildpack supports it? It is a bit of the chicken an egg problem, I guess. In my humble opinion, the minor version resolution should happen in the buildpack and not via some third party service, like the [nodejs buildpack](https://github.com/heroku/heroku-buildpack-nodejs/blob/b879e848921c815a973260067f05c26dbdf33001/lib/binaries.sh#L16) does.

freeformz:
@jtarchie If this lands before the buildpack changes they will be tied to the major go version until the BP can resolve the latest minor release (assuming they update their Godeps.json file and push). I don't like that behavior.

flavorjones:
@freeformz Is there a path forward you can recommend to add this support to Godep?

Alternatively, if we think the only consumers of `GoVersion` are the Heroku buildpack and the CloudFoundry buildpack (which is downstream from Heroku at the moment), perhaps we should just make this change in the buildpack and not touch this field at all?

freeformz:
@flavorjones #342 merged as v39.

flavorjones:
Thank you!

PR #321, opened 2015-11-10
Fixes related to symlinks and legal files

michaeltrobinson:
Fixed issue where Godep would think files weren't tracked by VCS if the GOPATH had a symlink.
Fixed issue where LICENSE files within dependencies vendor directories would be re-vendored, resulting in failing builds.

### DIFF
--- a/save.go
+++ b/save.go
@@ -267,11 +267,20 @@ func removeSrc(srcdir string, deps []Dependency) error {
 }
 
 func copySrc(dir string, deps []Dependency) error {
+	var err error
 	// mapping to see if we visited a parent directory already
 	visited := make(map[string]bool)
 	ok := true
 	for _, dep := range deps {
 		srcdir := filepath.Join(dep.ws, "src")
+		srcdir, err = filepath.EvalSymlinks(srcdir)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


save.go
```diff
@@ -316,7 +325,7 @@ func copySrc(dir string, deps []Dependency) error {
 		w = fs.Walk(rootdir)
 		for w.Step() {
 			fname := filepath.Base(w.Path())
-			if IsLegalFile(fname) {
+			if IsLegalFile(fname) && !strings.Contains(w.Path(), sep) {
```
freeformz: Can you add a test case for this?

freeformz:
It's probably best if you want to discuss these / work these through to split this PR into two. One for each issue.

PS: Symlinks inside $GOPATH aren't really supported (ex. https://groups.google.com/d/msg/golang-dev/GLUZGtUSj8A/SGUtCooLIT0J), so I'm curious as to how you are using them and why we would need to resolve them.

michaeltrobinson:
Okay, I will add a test case and split this into two PRs with more details

freeformz:
Thank you!

PR #322, opened 2015-11-11
Fix license vendoring

michaeltrobinson:
This resolves an issue I was seeing where Godep would vendor LICENSE files dependencies dependencies. 

I saw this specifically when importing a package that depended on the AWS SDK which depends on go-ini which contains a LICENSE file. I would then get an error from `go build` like `no buildable Go source files in my-project/vendor/github.com/aws/aws-sdk-go/vendor/github.com/vaughan0/go-ini` and `vendor/github.com/aws/aws-sdk-go/vendor/github.com/vaughan0/go-ini` only contains a LICENSE file. 

Added a test case that confirms this fix.

### DIFF
--- a/save.go
+++ b/save.go
@@ -316,7 +316,7 @@ func copySrc(dir string, deps []Dependency) error {
 		w = fs.Walk(rootdir)
 		for w.Step() {
 			fname := filepath.Base(w.Path())
-			if IsLegalFile(fname) {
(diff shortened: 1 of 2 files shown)


freeformz:
I started looking into this and I think this is really a side effect of a bug in /vendor/ (and possibly Godeps workspace handling or because of it). I'm digging in more.

freeformz:
Merged as 5e3d21abab679aef179d5dce308289619a59cf22

Thanks!

PR #326, opened 2015-11-13
fix isDirty command

### LINKED ISSUE #303: Godep not handling nested packages cleanly

azylman:
It's pretty conventional in Go land to have a repository that has multiple commands built into it. For example, see: https://github.com/kubernetes/kubernetes/tree/master/cmd

Sometimes, you might want to vendor these different commands separately, which would lead you wanting to have a vendor directory in each folder. Right now, godep doesn't support this because it's
"isDirty" check is overzealous. Instead of checking just the package that it takes as an argument, it checks the _whole_ repository that that package is in.

This commit changes it for git (I don't use other VCS and don't know if they work correctly or are also broken) so that it only checks the package that it's trying to vendor instead of the whole repository.

Fixes #303 and #305.

### DIFF
--- a/vcs.go
+++ b/vcs.go
@@ -40,7 +40,7 @@ var vcsGit = &VCS{
 
 	IdentifyCmd: "rev-parse HEAD",
 	DescribeCmd: "describe --tags",
-	DiffCmd:     "diff {rev}",
+	DiffCmd:     "diff {rev} --dirstat .",
 	ListCmd:     "ls-files --full-name",
 	RootCmd:     "rev-parse --show-toplevel",
 


freeformz:
I don't understand this use case.

I use and have several multi-command repositories that use godep.

Here is one of them: https://github.com/heroku/shh

When working with this project we `cd` into `$GOPATH/src/github.com/heroku/shh` and run `godep save -r ./...` and all of the deps for everything are vendored/rewritten.

azylman:
There's a couple reasons you might want this. For us, where this came up is that the top-level package is a library and it has a command-line interface that consumes it in `<pkd>/cmd/<binary name>`. We want to vendor dependencies in the cmd but not in the library, because other packages consume that library, so ideally we'd have `<pkg>/cmd/vendor`.

Another reason that I mentioned in the pull request is that you want different versions of the same dependency for different commands within your repository, which is less common, but definitely happens (e.g. at companies that are big into monorepos). I think this was theuse case mentioned in #303 and #305.

I think regardless of whether or not you agree with the use case, it should be pretty clear that the `isDirty` function right now is broken: it takes in a package and purports to tell you if that package is dirty, but instead it tells you if the repo that that package is in is dirty, which isn't always the same thing. Luckily, this can be fixed this while enabling new use cases without breaking existing use cases.

freeformz:
a) godep is meant to be run against the root of a repo containing one or more packages. If you have a structure of `./mylib/cmd/cmdname` and run godep save in the `cmdname` directory `mylib` will be vendored again in `cmdname's` Godep workspace. This will likely lead to additional confusion.

b) A PR that changes a single vcs command will not be accepted unless all vcs commands are updated, or proven to not need the same update. In the past this has led to problems/breakage in other vcs.

c) The change probably doesn't do what you want as `git diff <rev> .` still considers sub dirs of `.`. So if you are running this in a `cmd` directory it will detect the library as dirty if any of the sub dirs/packages are dirty. Or at least it does with my version of git (2.6.3) / local config. I googled around for a "fix", but I don't think there's a simple fix to this w/o filtering things like dirstat output.

azylman:
a) I'm aware that it's vendored again - I think this is less confusing than the current behavior

b) That's unfortunate

c) Yes, good point. In the use case that I was testing this on, it was actually sibling packages, not parent/child, so I didn't notice that. My guess is that there's not really a good way to do it for parent/child, since it would require knowledge of your project - e.g. when is a subpackage relevant and when it isn't. This would still fix the monorepo sibling packages case mentioned in #303 and #305

freeformz:
FWIW: I think the point of isDirty(dir,rev) was to determine if the repo root is dirty, not any given directory in a repo, but I'd have to let @kr chime in.

kr:
Yeah, that was the original intention. But if some other behavior would be more useful, it's not set in stone. :)

azylman:
Is there any risk to vendoring a package in a dirty repository when the package itself isn't dirty? I can't think of anything, but might be missing something.

Obviously you don't want to vendor a dirty package because then the code doesn't correspond to any commit.

freeformz:
@azylman if that unmodified package pulls in other packages in the same repo that aren dirty and have local changes that you rely on? I guess we'd pick that up as a dependency on bomb out there though. So no, I don't think there's any problem with doing that, code just need to be adjusted properly.

freeformz:
FWIW This only works with git, what about the other vcs? Whatever we do we need to be able to support it across all supported vcs implementations.

freeformz:

PR #328, opened 2015-11-17
Get Test Package Dependencies Option

bbokorney:
This feature adds a `-t` flag to the `godep get` command to download and install test package dependencies. This is similar to the `-t` flag in the `go get` command.

I wasn't sure how to handle the version change so I simply incremented the version number. Please let me know if this should be done differently.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+# v30 2015/11/25
+
+* Add `-t` flag to the `godep get` command.
+
 # v29 2015/11/17
 
 * Temp work around to fix issue with LICENSE files.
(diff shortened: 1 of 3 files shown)


bbokorney:
Any feedback on this change?

freeformz:
@bbokorney Since you are updating the version, please also update the Changelog and squash down to a single commit. Thanks!

bbokorney:
Okay, I'll do that right now.

bbokorney:
Alright, I've updated the Changelog.

freeformz:
Thanks!

PR #330, opened 2015-11-19
In restore errors, mention which package had the problem.

shurcooL:
This improves the situation when the user runs `godep restore` and sees a few instances of "restore, during download dep" errors. Previously, it was not mentioned which packages failed to download. Turning on -v verbose mode was not a good solution since it prints too much other information for successfully processed packages.

### DIFF
--- a/restore.go
+++ b/restore.go
@@ -29,7 +29,7 @@ func runRestore(cmd *Command, args []string) {
 	for _, dep := range g.Deps {
 		err := download(dep)
 		if err != nil {
-			log.Println("restore, during download dep:", err)
+			log.Printf("restore, during download dep %q: %v\n", dep.ImportPath, err)
 			hadError = true
 		}
 	}
@@ -37,7 +37,7 @@ func runRestore(cmd *Command, args []string) {
 		for _, dep := range g.Deps {
 			err := restore(dep)
 			if err != nil {
-				log.Println("restore, during restore dep:", err)
+				log.Printf("restore, during restore dep %q: %v\n", dep.ImportPath, err)
 				hadError = true
 			}
 		}


shurcooL:
Any feedback on this change?

drewwells:
Hmm, what error were you getting from `git|othervcs`?

For a git submodule error, I'm only getting exit status 1. While it lets me know which package is problematic, it's not detailed enough to know what happened. Here's the comparison with verbose. Maybe on errors, it should report stdout and stderr?

```
-> % godep restore ./...
godep: restore, during download dep "github.com/wellington/go-libsass": exit status 1
``` 

Here's the detailed error
```
 output: github.com/wellington/go-libsass (download)
# cd .; git --git-dir=/Users/drew/src/github.com/wellington/go-libsass/.git submodule update --init --recursive
No submodule mapping found in .gitmodules for path 'libsass-src'
package github.com/wellington/go-libsass: exit status 1
```

shurcooL:
I don't remember what the underlying problem was exactly when I ran into this. I think either the repository was deleted and no longer existed upstream. Or perhaps just the commit wasn't found and that produced the error.

The main problem was that I had no way of knowing what package even had the problem. This small change resolves that.

I agree that it'd be nicer still to decorate the error returned from `download` to include more info on the error (such as stderr), but I think that can be a separate PR.

shurcooL:
I think this PR can/should be merged since it's a strict improvement over the previous behavior, where it was completely impossible to know which package had download/restore problems.

Any reason not to?

freeformz:
@shurcooL Thanks, merged as cf33bd75a8243ac701090bb6fa8e0b81b9746fcb

PR #336, opened 2015-11-26
Vendor applications safely

krnowak:
For go 1.4, adding a dummy go file with some import clauses importing
"main" packages was enough for godep to vendor them. With go 1.5 it is
not working anymore, because go will bail out when loading a package
with this dummy go file with an error 'import 'some/main/package' is a
program, not an importable package'.

This introduces reading the "appVendor" file, which is basically a
list of main packages to import. It is not a go source file, so it
sidesteps the issue with the new go.

The reason to add this is that the project may want to use some
application from specific revision during building or testing. git
submodules allow that, but the problem can arise when the project
disappears. "go get" does not allow getting a specific revision.

The above is a commit message, so a small addition follows. The specific use case is in [coreos/rkt](https://github.com/coreos/rkt), where we have [a dummy.go file](https://github.com/coreos/rkt/blob/master/stage1/dummy.go) importing some "main" packages which are built by our build system. This worked fine when we were using go 1.4, but it stopped working with go 1.5.

I'm not sure about two points in this PR:

- The location and filename of the file, currently we expect it to be in project's root (like `Godeps` directory) and have a name `appVendor`.
  - I suppose it would be better if it was inside `Godeps` directory, but:
    - It would require the `godep` user to create the directory.
    - It would require `godep` to recreate the file as whole `Godeps` directory gets removed during `save` command (right?). Not a problem per se, just more code to be written.
  - The `appVendor` name is… meh. But naming is hard.
  - I decided against another JSON file, because we might want to have comments there.
- Should we allow both styles of comments (script like `#` and C-or-go-like `//`)?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -54,6 +54,29 @@ command, you wrap it in one of these two ways:
 
 Test files and testdata directories can be saved by adding `-t`.
 
+It is also possible to vendor "main" packages, but not through a go source file
+in the project. Instead, create a file named "appVendor" and add the desired
+packages there, one in each line. The file may look like as follows:
+
+```
+// You can use double slash to write a comment
+# But you can also use hash for the same purpose
+
+github.com/fizz/buzz
+github.com/foo/bar
+
+
+// Empty lines are also allowed
+github.com/one/two // Comments after package are allowed
+github.com/three/four # Same with hash comments
+
+// Vendor in the quux app, we build it ourselves and use it in tests
+github.com/a/b/apps/quux
+```
+
+All the packages in "appVendor" file together with their dependencies will be
+added to the Godeps file.
+
 ## Additional Operations
 
 ### Restore
--- a/appvendor_test.go
+++ b/appvendor_test.go
@@ -1,0 +1,32 @@
+package main
+
+import (
+	"reflect"
+	"strings"
+	"testing"
+)
+
+func TestAppPkgs(t *testing.T) {
+	testAppVendorContents := `
+// some slash comment
+# some hash comment
+
+github.com/tools/godep
+github.com/golang/go // package with a slash comment
+github.com/golang/tools # package with hash comment
+
+`
+	expectedPkgs := []string{
+		"github.com/tools/godep",
+		"github.com/golang/go",
+		"github.com/golang/tools",
+	}
+	r := strings.NewReader(testAppVendorContents)
+	pkgs, err := getAppPkgsFromReader(r)
+	if err != nil {
+		t.Errorf("Failed to parse %q: %v", testAppVendorContents, err)
+	}
+	if !reflect.DeepEqual(pkgs, expectedPkgs) {
+		t.Errorf("App pkgs = %v want %v", pkgs, expectedPkgs)
+	}
+}
--- a/appvendorfile.go
+++ b/appvendorfile.go
@@ -1,0 +1,43 @@
+package main
+
+import (
+	"bufio"
+	"io"
+	"os"
+	"strings"
+)
+
+var (
+	appVendorFile = "appVendor"
+)
+
+func getAppPkgsFromReader(r io.Reader) ([]string, error) {
+	s := bufio.NewScanner(r)
+	var pkgs []string
+	for s.Scan() {
+		l := s.Text()
+		l = strings.SplitN(l, "#", 2)[0]
+		l = strings.SplitN(l, "//", 2)[0]
+		l = strings.TrimSpace(l)
+		if l == "" {
+			continue
+		}
+		pkgs = append(pkgs, l)
+	}
+	if err := s.Err(); err != nil {
+		return nil, err
+	}
+	return pkgs, nil
+}
+
+func getAppPkgs() ([]string, error) {
+	f, err := os.Open(appVendorFile)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	defer f.Close()
+	return getAppPkgsFromReader(f)
+}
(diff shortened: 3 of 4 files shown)


freeformz:
godep save supports saving any package specified in the pkg spec.
We are using that internally to vendor `github.com/zimbatm/socketmaster` into an application. socketmaster is a `main` package.

So specifying something like `godep save -r ./... github.com/zimbatm/socketmaster` works as expected. using something like `jq` the heroku buildpack [reads the Packages out of the Godeps/Godeps.json](https://github.com/heroku/heroku-buildpack-go/blob/52ebcad122e4cb12ae9bf9bba13aa9f0342fd5d6/bin/compile#L177) file and installs all of the packages specified there with [`godep go install $pkgs`](https://github.com/heroku/heroku-buildpack-go/blob/52ebcad122e4cb12ae9bf9bba13aa9f0342fd5d6/bin/compile#L227).

Could you try something like that instead?

freeformz:
PS: I'm also -1 wrt adding another file.

jonboulle:
Hmm interesting, so rather than relying on godep's automagical import detection, we should always be explicitly listing main packages when we invoke?

freeformz:
@jonboulle the package spec passed to `godep save` is what feeds godep the list of packages to inspect for 'automagical detection'. I would only list `main` packages if you want them saved. Arguably `godep save` w/o any args should default to `.` or the contents of the `Godeps/Godeps.json` Packages array, if the file exists. This would probably help a lot and help people DTRT more often. Thoughts?

jonboulle:
> @jonboulle the package spec passed to godep save is what feeds godep the list of packages to inspect for 'automagical detection'.

Right, that actually was my understanding, which is why I was surprised at your suggestion (which I misinterpreted). The thing is, `godep save -r ./... github.com/zimbatm/socketmaster` only partly does what I expect/want; it will a) vendor+rewrite all the package imports it discovers under all the packages in `./...` (great), and b) vendor+rewrite `github.com/zimbatm/socketmaster` into my current `$PWD/Godeps/` directory (also great). BUT, it also c) rewrites all of the imports WITHIN `$GOPATH/src/github.com/zimbatm/socketmaster`, which is definitely NOT what I want. Am I doing something wrong, or misunderstanding this yet again?

jonboulle:
> Arguably godep save w/o any args should default to . or the contents of the Godeps/Godeps.json Packages array, if the file exists. This would probably help a lot and help people DTRT more often. Thoughts?

Hmmm are you suggesting that `Godeps/Godeps.json` now becomes an editable source manifest rather than something only written to by the `godep` tool? That is arguably much more intuitive (in my experience it's what every newcomer to godep EXPECTS to happen when they see the manifest), but totally inverting the current workflow right?

freeformz:
Re Rewrites: it would rewrite and vendor any of the imports the vendored tool imports. But you'de want that wouldn't you? That's the only way to guarantee you can build that tool later.

re: Godeps.jon: It's already a manifest, and I don't mean manually editable. Yes, it changes the current workflow a bit (but not much). I'm basically saying that every time you do a `godep save` it reads in `Godeps/Godeps.json` and if there is a Packages array there it uses those values for the package spec. If there is none, it defaults to `.`. If you specify a new package spec with a save it uses that one, overwriting the existing one in the file. This is where problems could come up though.

jonboulle:
On Wed, Dec 2, 2015 at 3:54 PM, Edward Muller <notifications@github.com>
wrote:

> Re Rewrites: it would rewrite and vendor any of the imports the vendored
> tool imports. But you'de want that wouldn't you? That's the only way to
> guarantee you can build that tool later.
>
>
I only want them rewritten within the _vendored_ version of the tool
imports; but godep is reaching outside my current working directory and
rewriting them in their $GOPATH source directories as well. Incidentally
that breaks subsequent `godep save` calls on those imports because then
their git working trees are dirty.

freeformz:
Ahh yeah. I missed that. For instance github.com/zimbatm/socketmaster doesn't have any external deps, so no rewrites. You could *not* use rewrites for now, but either way that's a bug,

jonboulle:
oh, good, so you'd take a fix? :-)
At this stage I am pretty sure that would solve our use cases /cc @dgonyeo

freeformz:
Hmm. This may break how some people use godep, not sure. The way *I'd* like godep used is that it would only rewrite packages that are specified via a relative package spec (`., ./, ./..., ./cmd/..., etc`) and their dependencies. In my mind it should not rewrite other package specs at their original location(s). But I can see how some people may use the fully qualified package spec when doing save (e.g.: `godep save github.com/me/thepackageImworkingon`). I've also seen people construct oddities in Makefiles to enumerate packages and pass the fully qualified package names to godep (I don't know *why* they do that though).

This would also require some tests, if they don't exist.

jonboulle:
On Wed, Dec 2, 2015 at 4:21 PM, Edward Muller <notifications@github.com>
wrote:

> Hmm. This may break how some people use godep, not sure. The way *I'd*
> like godep used is that it would only rewrite packages that are specified
> via a relative package spec (., ./, ./..., ./cmd/..., etc) and their
> dependencies. In my mind it should not rewrite other package specs at their
> original location(s).
>
+1, agree with all these points

> But I can see how some people may use the fully qualified package spec
> when doing save (e.g.: godep save github.com/me/thepackageImworkingon).
> I've also seen people construct oddities in Makefiles to enumerate packages
> and pass the fully qualified package names to godep (I don't know *why*
> they do that though).
>
Hmm, yeah. I seem to recall there being subtle but nefarious differences in
behaviour between these approaches which caused us to end up very firmly on
our current ./... invocation. But the details escape me.

freeformz:
If you want to take a crack at any of the things we discussed I'd be happy to review. For now I'm going to close this PR though. Thanks!

jingweno:
Hmm...didn't know `godep save -r ./.. PKG` will rewrite `PKG` in `$GOPATH/src/PKG`. Seems like a we should fix this for these reasons:

1) People wouldn't be interested in rewriting import paths outside of their project
2) An intuitive interpretation of `godep save -r PKG` is pulling PKG to the project and rewrite its import paths, instead of rewriting the import paths of  `$GOPATH/src/PKG`.

Issue #271, opened 2015-09-04
Vendor everything regardless of build tags


freeformz:
Currently we end up respecting build tags when we select files/packages for vendoring. This causes problems when vendoring happened on a target platform that isn't the one you are building on. We should ignore build tags when selecting files.

captncraig:
I have an app that we build for multiple os/arch combinations. godep is not working for me because it only includes files for the current build tags, and deletes previous ones from other build tags. I would like to switch to godep, but it is difficult if this is not resolved.

captncraig:
It looks like this will be difficult to do with the current implementation that executes `go list`, as that has no option to ignore tags. It should be feasible by invoking `go/build` things directly. I may make an attempt in that direction.

captncraig:
On second thought, it may be difficult to do with `go/build` too. You can import a package using `UseAllFiles`, which will ignore all tags, but in my testing that tends to expose a lot of errors with packages that have a secondary main package with `// +build ignore`, or things like that.

`build.Package` does expose `AllTags` which tells you which tags are relevant in any way. You could potentially use that to call Import multiple times, but it is tricky to know which values are for GOOS, which are for GOOARCH, and which are to be given as raw tags. Not to mention things like ignore or race should be special cased. Build constraints on a file may always be in an odd combination like `// +build !linux,amd64` in which case I don't have a clue what to do.

The real problem I have is that running `GOOS=windows godep save` deletes packages not needed for the windows build. If I then run `GOOS=linux godep save`, it will add linux dependencies, but delete the windows ones. Perhaps this behavior could be disabled by a flag and print a warning or something like that so I can manually prune things if needed. That may be a simpler solution.

darkliquid:
I get bitten by this issue a lot as well. One thing I've been toying with (but have yet to actually write) is a script to that will take a list of GOARCH and GOOS that your project supports, then run godep against every combination, at each step unstaging any deletions and adding any new packages (assuming git). It a right pain in a backside to have to do this though - it'd be nice if go list had some kind of 'ignore all tags' option that would make it list everything all the time (maybe my passing in \* to the -tags argument). Until then, manual hacks and workarounds seem to be the only way forward.

freeformz:
@captncraig Yep. I've spent a lot of time going down that same path and before I went on vacation (which I'm back from) I started writing my own version of go/build.Context because of this. It's no where near ready though. ;-)

ecnahc515:
[...]

PR #337, opened 2015-11-27
Add -a flag to vendor regardless of build tags

gyuho:
This adds a flag `-a` when people need to vendor everything regardless of build
tags. This fixes https://github.com/tools/godep/issues/271.

### DIFF
--- a/pkg.go
+++ b/pkg.go
@@ -2,9 +2,16 @@ package main
 
 import (
 	"encoding/json"
+	"go/build"
+	"go/parser"
+	"go/token"
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"sort"
+	"strings"
+	"sync"
 )
 
 // Package represents a Go package.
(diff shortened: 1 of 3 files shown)


gyuho:
@freeformz This might be a little hacky but I tried it and it worked.

Easy example is https://github.com/boltdb/bolt. It has only one external dependency of `golang.org/x/sys/unix` in https://github.com/boltdb/bolt/blob/master/bolt_unix_solaris.go. If you don't `godep` in Solaris, it will spin out empty `Godeps.json`.

I cannot think of an easier way because `go list` command gets its list from `go build` command, which is very hard to change. This approach would give us what we want, I hope.

Please let me know if you have any feedback.

gyuho:
Nvm. This might not work for projects with existing Godeps. Will investigate more. :(

freeformz:
LoadPackages still uses `go list`. I opened a PR yesterday that removes `go list` and also solves the missing deps problem: #341, effectively making your `-a` just the way godeps works (at least AFAICT).

gyuho:
Great. Thanks @freeformz . I will close this in favor of https://github.com/tools/godep/pull/341/.

PR #338, opened 2015-11-27
Dependencies sorting

JekaMas:
Hello!
As i can see godep now has problem that after saving dependencies after few commits, you can get very different file .json. What creates great confusion when merge, especially in large projects: a big diff of .json file is difficult to understand and difficult to figure out whether developer change a lot of dependencies, or only one or two, but godep changed their order when saving a file.
I think that the addition of sorting dependencies would be useful.

### DIFF
--- a/dep.go
+++ b/dep.go
@@ -81,3 +81,21 @@ func goVersion() (string, error) {
 	}
 	return p[2], nil
 }
+
+// dependencySlice - slice of Dependency
+type dependencySlice []Dependency
+
+// Len implements sort Len interface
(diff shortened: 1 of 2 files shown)


freeformz:
FWIW: I've never run into this and use godep almost daily. I'm not against sorting the dependencies though. I would like a repeatable example of when/how this happens and a test for it.

freeformz:
@JekaMas Can you provide an example of this workflow?

freeformz:

Issue #334, opened 2015-11-25
GOPATH as symlink confuses godep


rrh:
I had my GOPATH environment variable pointing to a sym link; the sym link pointed to my current go development tree.  This situation caused godep save to decide that new packages to vendor contained nothing worth saving, presumably because the symlink introduced aliases that were not correctly accounted for.

I spent 10+ hours debugging this, and at one poing gave up on godeps entirely.

kr:
I'm sorry for this. It's a good idea to avoid using symlinks for elements of GOPATH, or inside your workspace.

The go tool includes code to explicitly prohibit some cases, though it seems there are still others that it doesn't catch. We (godep) should consider taking a similar approach, to find cases that are unlikely to work correctly and produce a helpful error message instead of silently doing something useless.

freeformz:
Closing issues before archiving. Thank You\!

PR #340, opened 2015-12-01
EvalSymlinks in Contains() check

### LINKED ISSUE #339: godep save thinks everything's unversioned
I'm certain this is a misconfiguration from go convention, and i'm too blame, but I can't figure it out.

```go list -json``` returns what looks valid:

https://gist.github.com/imajes/8ae3025edb35c140446d

but, ```go save -v ./...``` seems to disagree:

https://gist.github.com/imajes/412835d9393eac768faf

and creates ```Godeps/``` with a completely empty ```_workspace```.

would anyone be able to help me figure out what's going wrong?

freeformz:
git's root command (maybe others vcs as well) resolve symlinks, so try
that too.

PS: In git's case we can possible use rev-parse --show-cdup + extra
logic to find the root of the repo instead. Need to figure out what
other vcs root commands do as well.

Fixes #339

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,8 @@
-# v31 2015/12/2
+# v32 2015/12/02
+
+* Eval Symlinks in Contains() check.
+
+# v31 2015/12/02
 
 * In restore, mention which package had the problem -- @shurcool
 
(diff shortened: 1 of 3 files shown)


freeformz:
This may also fix #334.

PR #341, opened 2015-12-01
Replace `go list`

### LINKED ISSUE #271: Vendor everything regardless of build tags

freeformz:
Main motivation is so that we can  pick up packages across architectures and to have more control over the list of packages.

I consider this the first in several structural changes.

Most additions are inside of list.go.

For each run of listPackage a depScanner is made to track already seen packages (inspired by Scanner). That list is not cached between executions.

The current implementation of listPackage attempts to resolve vendor/ a bit better than we used to, but other changes need to be implemented elsewhere in godep before that can be fully realized.

All *existing* tests pass, but this doesn't mean it actually works *better* (yet).

/cc @kr @neurogeek 

Fixes #271


kr:
I think this is the right approach in the long run. It had to happen eventually.

pkg.go
```diff
@@ -27,46 +37,129 @@ type Package struct {
...
+	var imports, testImports []string
+	for _, fname := range p.IgnoredGoFiles {
+		tgt := filepath.Join(p.Dir, fname)
+		fset := token.NewFileSet()
+		f, err := parser.ParseFile(fset, tgt, nil, parser.ParseComments)
+		if err != nil {
+			return nil, nil, err
+		}
+		if len(f.Comments) > 0 {
+			for _, c := range f.Comments {
```
kr: Looks like this scans all comments, but build tags have to appear near the top of the file. Might get false positives from looking at too many comments.

Also… is this necessary at all? Why not include "ignore" files too for the purposes of dependencies? Ignore doesn't necessarily mean "this file is garbage", it simply means "don't build this file into the package", same as any other unsatisfied constraint.

pkg.go
```diff
@@ -27,46 +37,129 @@ type Package struct {
...
+				ct := c.Text()
+				if i := strings.Index(ct, buildMatch); i != -1 {
+					for _, b := range strings.FieldsFunc(ct[i+len(buildMatch):], buildFieldSplit) {
+						if b == "ignore" {
+							continue
```
kr: This doesn't seem to accomplish anything. The body of this loop is otherwise empty. Does there need to be a label on the outermost loop?

pkg.go
```diff
@@ -27,46 +37,129 @@ type Package struct {
...
+	}
+	var imports, testImports []string
+	for _, fname := range p.IgnoredGoFiles {
+		tgt := filepath.Join(p.Dir, fname)
+		fset := token.NewFileSet()
+		f, err := parser.ParseFile(fset, tgt, nil, parser.ParseComments)
+		if err != nil {
+			return nil, nil, err
+		}
+		if len(f.Comments) > 0 {
```
kr: This if statement looks redundant; if there are no comments, the for loop will run zero times.

pkg.go
freeformz: It's not necessary, at least not now. This was left over from a different experiment. I've removed this method.

pkg.go
freeformz: removed unused method.

pkg.go
freeformz: removed un-used method.

freeformz:
@kr sorry about the noise a bunch of stuff in pkg.go slipped in from another, earlier experiment.

freeformz:
Would close #337

PR #342, opened 2015-12-06
Record / Update only the major go version

freeformz:
The idea is that we shouldn't care about the minor revision of Go.

/cc @kr, @neurogeek

### DIFF
--- a/dep.go
+++ b/dep.go
@@ -68,8 +68,17 @@ func uniq(a []string) []string {
 	return a[:i]
 }
 
-// goVersion returns the version string of the Go compiler
-// currently installed, e.g. "go1.1rc3".
+// trimGoVersion and return the major version
+func trimGoVersion(version string) (string, error) {
+	p := strings.Split(version, ".")
+	if len(p) < 2 {
+		return "", fmt.Errorf("Error determing major go version from: %q", version)
+	}
+	return p[0] + "." + p[1], nil
+}
+
+// goVersion returns the major version string of the Go compiler
+// currently installed, e.g. "go1.5".
 func goVersion() (string, error) {
 	// Godep might have been compiled with a different
 	// version, so we can't just use runtime.Version here.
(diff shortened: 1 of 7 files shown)


update.go
```diff
@@ -12,28 +13,64 @@ import (
...
+)
 
 func init() {
 	cmdUpdate.Flag.BoolVar(&saveT, "t", false, "save test files during update")
+	cmdUpdate.Flag.BoolVar(&updateGoVer, "goversion", false, "update the recorded go version")
 }
 
 func runUpdate(cmd *Command, args []string) {
-	err := update(args)
+	if updateGoVer {
```
kr: Nice.

dep.go
```diff
@@ -64,8 +64,17 @@ func uniq(a []string) []string {
...
 
-// goVersion returns the version string of the Go compiler
-// currently installed, e.g. "go1.1rc3".
+// trimGoVersion and return the major version
+func trimGoVersion(version string) (string, error) {
+	p := strings.Split(version, ".")
+	if len(p) < 2 {
+		return "", fmt.Errorf("Error determing major go version from: %q", version)
+	}
+	return strings.Join([]string{p[0], p[1]}, "."), nil
```
kr: Why not `p[0] + "." + p[1]`?

kr:
LGTM

neurogeek:
LGTM too.

I would like to know what prompted this? Just clean up or is there a more 'profound' reason?
Thanks!

dep.go
freeformz: Heh. No reason. That's better, I'll change it.

freeformz:
@neurogeek After some issue discussion (heroku/heroku-buildpack-go/100 and another I can't find atm) I generally think that we should really only care about the major version of Go and let downstream systems care about selecting the minor version (the user or build systems). There are probably exceptions to that, hence I also stopped auto-updating the go version on save, allowing people to specify whatever they want if they manually want to manage the file.

neurogeek:
Makes sense. 
Thanks for the details.

flavorjones:
Awesome, thanks for this!

PR #343, opened 2015-12-07
Rewrite packages only inside the project

krnowak:
When calling "godep -r ./... github.com/example/proj", godep will
import the example/proj and its dependencies into our project tree and
will rewrite the paths accordingly. But it will also modify the
example/proj repository in GOPATH. In other words - the example/proj
modified in two places - in source directory (in
$GOPATH/src/github.com/example/proj) and in destination directory (in
./Godeps/_workspace/src/github.com/example/proj).

This commit fixes it, so the packages outside the cwd are not
rewritten.

### DIFF
--- a/save.go
+++ b/save.go
@@ -48,9 +48,10 @@ The dependency list is a JSON document with the following structure:
 Any packages already present in the list will be left unchanged.
 To update a dependency to a newer revision, use 'godep update'.
 
-If -r is given, import statements will be rewritten to refer
-directly to the copied source code. This is not compatible with the
-vendor experiment.
+If -r is given, import statements will be rewritten to refer directly
+to the copied source code. This is not compatible with the vendor
+experiment. Note that this will not rewrite the statements in the
+files outside the project.
 
 If -t is given, test files (*_test.go files + testdata directories) are
 also saved.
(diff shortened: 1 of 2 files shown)


krnowak:
Oh, and this PR is related to comments in #336.

freeformz:
Thanks for taking the time to write this PR. FWIW: I'm -1 another flag. I'm +1 for an implementation that skips rewrites for any package specified that falls outside of the package in the cwd and it's sub packages though. While this may nerf godep for some people I think it's a what most people would expect and the least surprising.

krnowak:
I had some problems after rebasing. I solved them with #346 and #351. So my further work would depend on those. Will put something tomorrow.

krnowak:
Forgot to mention that I updated this PR.

jonboulle:
@krnowak For posterity I suggest updating the wording of the PR OP to match the new commit message

krnowak:
@jonboulle: How about now?

jonboulle:
@krnowak I meant [here](https://github.com/tools/godep/pull/343#issue-120770486)

krnowak:
@jonboulle: Done.

jonboulle:
LGTM

jonboulle:
@krnowak time for a rebase
@freeformz ping on this?

freeformz:
It probably needs another rebase. Let me take a quick look and make some code comments too.

save.go
```diff
@@ -95,6 +96,22 @@ func dotPackage() (*Package, error) {
 	return p[0], nil
 }
 
+func projectPackages(dot *Package, a []*Package) []*Package {
```
freeformz: recent changes have done away with the dot package reference because we were only using the dir anyway.

save.go
```diff
@@ -95,6 +96,22 @@ func dotPackage() (*Package, error) {
...
+	var projPkgs []*Package
+	dDir, err := filepath.EvalSymlinks(dot.Dir)
+	if err != nil {
+		dDir = dot.Dir
+	}
+	dotDir := fmt.Sprintf("%s/", dDir)
```
freeformz: Probably want `dDir + filepath.Separator`

save.go
```diff
@@ -95,6 +96,22 @@ func dotPackage() (*Package, error) {
...
+	for _, p := range a {
+		pkgDir := fmt.Sprintf("%s/", p.Dir)
```
freeformz: ditto wrt seperator.

freeformz:
See comments, also a rebase (sorry). Thanks for adding the positions to the comments. Otherwise LGTM.

krnowak:
I brought the `dotPackage` function back, but now it returns our own small struct with only an import path and a directory. That way `projectPackages` function does not have to return an error too (from `filepath.Abs(".")`).

jonboulle:
bump :-)

krnowak:
Rebased to fix a minor conflict.

save.go
```diff
@@ -81,21 +82,46 @@ func runSave(cmd *Command, args []string) {
 	}
 }
 
-func dotPackageImportPath() (string, error) {
+type dotPkg struct {
+	importPath string
+	dir        string
+}
+
+func dotPackage() (*dotPkg, error) {
```
freeformz: We could just simple return a build.Package instead. I think I'd rather do that then add yet another "package" type to the code.

freeformz:
Small comment about the new type, otherwise LGTM

krnowak:
Updated, no more `dotPkg` type.

freeformz:
@krnowak Thanks for your patience with the extended review!

PR #346, opened 2015-12-08
Fix loading a dot package

krnowak:
Getting the dot package should succeed, even in spite of no buildable
Go sources in the directory. The only thing we need from the package
is the import path and that was provided even by `go list -e -json .`.

If the toplevel directory has no buildable Go sources and we invoke
godep save without any args, then it will fail later. For such
projects an arg like ./... should be passed.

This is a regression that appeared after the `go list ` replacement has landed.

### DIFF
--- a/pkg.go
+++ b/pkg.go
@@ -43,10 +43,12 @@ func LoadPackages(names ...string) (a []*Package, err error) {
 	}
 	for _, i := range importPaths(names) {
 		p, err := listPackage(i)
-		if err != nil {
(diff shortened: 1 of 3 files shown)


freeformz:
Thanks for the PR, but I took a different approach in #352.

indykish:
Thanks for this fix, we were scratching our head  last nite when our tree https://github.com/megamsys/megdc.git kept saying "no buildable source files".

I can save "Godeps" now.

Issue #345, opened 2015-12-08
Bug in godep restore


SDkie:
https://github.com/tools/godep/commit/bd1a5cd1a59bff09fa53022436490eb03fb0dc4c

Above commit has some issue. 'go dep restore' is giving me following error : "godep: error restoring dep (github.com/chai2010/webp): found packages webp (bench_test.go) and main (gen_helper.go) in /go/src/github.com/chai2010/webp"

Before that commit it was working properly.

krnowak:
I tried to fix this, but it is usually the package that needs to be fixed actually. Normally, you should not have two files in a directory with two different package names (like `bench_test.go` has a package `webp` and `gen_helper.go` has a package `main`).

This is a fallout from moving away from `go list` in favor of `go/build` package.

Usually, this kind of bug is because that often helper binaries have "main" package and a `// +build ignore` constraint. The constraint was enough for `go list` to ignore the file alright. But after the switch, the [`UseAllFiles`](https://golang.org/pkg/go/build/#Context) field is set to true by default, so it takes all the files, not taking `// +build` constraints into account at all.

Probably the proper solution would be to fix the offending package, but I tried to somehow work it around and failed.

freeformz:
"Fixed" in #352 as we'll also ignore the "ignore" build tag, at least for now. I think that's the path of least surprise.

PR #347, opened 2015-12-08
Just warn about multiple packages

krnowak:
There are many projects outside that have some helper binaries with a
"// +build ignore" clause and "main" package in a directory, where the
rest of files have some other package name.

So far I found [cznic/ql](https://github.com/cznic/ql/pull/114), [coreos/rkt](https://github.com/coreos/rkt/pull/1848) and golang.org/x/net/html/atom (found packages atom (atom.go) and main (gen.go) in /home/kv/projects/coreos/GOPATH/src/golang.org/x/net/html/atom)

This depends on #346, as both modify the code around the same place.

### DIFF
--- a/pkg.go
+++ b/pkg.go
@@ -43,10 +43,12 @@ func LoadPackages(names ...string) (a []*Package, err error) {
 	}
 	for _, i := range importPaths(names) {
 		p, err := listPackage(i)
-		if err != nil {
-			return nil, err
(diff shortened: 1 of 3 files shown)


krnowak:
Closing this. It is not enough and I have a problem fixing it. Will file an issue.

krnowak:
That's actually already filed - #345

PR #351, opened 2015-12-08
stow away helper files

krnowak:
Stow away the helper files when trying to import the package …

There are many projects outside that have some helper binaries with a
"// +build ignore" clause and "main" package in a directory, where the
rest of files have some other package name.

In this commit we are renaming them, so they won't get picked up by
build context. After we are done, they are renamed back.

This is just a POC addressing #345.

### DIFF
--- a/list.go
+++ b/list.go
@@ -3,6 +3,9 @@ package main
 import (
 	"fmt"
 	"go/build"
+	"go/parser"
+	"go/token"
+	"io/ioutil"
 	"log"
 	"os"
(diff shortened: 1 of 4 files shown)


krnowak:
Oh yeah, it depends on #346.

krnowak:
And sorry about the lack of the tests, will add them tomorrow if needed.

freeformz:
Thanks for the PR, but I took a different approach in #352.

PR #357, opened 2015-12-11
[WIP] Performance Regression Patch. First pass

### LINKED ISSUE #356: Performance regression
From @jonboulle comment in #352
...
This change appears to have had a pathological effect on performance in at least one of our repos (coreos/rkt):

```console
$ time /tmp/gopath/bin/godep32 save -r ./...
real    0m1.649s
user    0m1.383s
sys 0m0.341s

$ time /tmp/gopath/bin/godep33 save -r ./...
godep: found packages keygen.go (main) and keymap.go (keystoretest) in /tmp/gopath/src/github.com/coreos/rkt/pkg/keystore/keystoretest
real    0m1.729s
user    0m1.487s
sys 0m0.240s

$ time /tmp/gopath/bin/godep34 save -r ./...
real    0m41.054s
user    0m37.493s
sys 0m4.088s
```

freeformz:
Locally these changes are about a 50% time reduction for a complex test
repo. There's plenty more to do though.

- Enable cpu profiling via flag on save.
- do a better compare in depScanner.Add(), comparing only the path /
  import instead of the entire struct
- remove checkGoroot as it was unused
- return if package was found to be a GOROOT package early as there's
  not much to do.
- Only create a new packageContext if we're actually adding it.
- Don't do a full scan of the dot package as we were only using it's import
  path anyway.

Fixes #356 (partially)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,12 @@
+# v35 2015/12/11
+
+* Fixes #356: Major performance regressions in v34
+    * Enable cpu profiling via flag on save.
+    * Cache packages by dir
+    * Don't do a full import pass on deps for packages in the GOROOT
+    * create a bit less garbage at times
+* Generalize -v & -d flags
+
 # v34 2015/12/08
 
 * We now use build.Context to help locate packages only and do our own parsing (via go/ast).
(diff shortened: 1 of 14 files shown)


freeformz:
My main performance testing repo https://github.com/coreos/rkt  went from a little less than 60s to a little less than 4s on my laptop.

PR #362, opened 2015-12-15
Change up how download/restore works a little

freeformz:
* We now try to load the package after downloading/restoring. Previously
  we did that too early in the process.
* make previous verbose output debug output
* report a typed error instead of a string from listPackage so we can
  assert and provide a nicer error.
* Catch go get errors that say there are no go files found. See code
  comment as to why.
* do *all* downloading during download phase.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,15 @@
+# v37 2015/12/15
+
+* Change up how download/restore works a little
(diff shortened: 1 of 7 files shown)


eparis:
I tested this change + https://github.com/kubernetes/kubernetes/pull/18706 and kube is once again happy!

PR #378, opened 2015-12-22
Fix a bunch of GO15VENDOREXPERIMENT issues

### LINKED ISSUE #358: godep save with GO15VENDOREXPERIMENT V35
```bash
$ export GO15VENDOREXPERIMENT=1
$ mkdir -p $GOPATH/src/github.com/test/pkg
$ cd $GOPATH/src/github.com/test/pkg
$ cat > main.go <<EOF
package main

import "github.com/bitly/go-simplejson"

func main() {
	simplejson.New()
}
EOF
$ go get
$ godep save
```

`vendor` directory is not created and `simplejson` is missing in `Godeps/Godeps.json`:
```json
{
	"ImportPath": "github.com/eleme/test",
	"GoVersion": "go1.5.1",
	"Deps": []
}
```

`V34` works as expected, am I expecting the wrong thing?

freeformz:
- Find package directories better. Previously we used build.FindOnly
  which didn't work the way I expected it to (any dir would work w/o
  error).
- Set the VendorExperiment bool based on go version as 1.6 defaults to
  on.
- A bunch of extra debugging for use while sanity checking myself.
- vendor flag for test structs.
- Some tests for vendor/ stuff:
    - Basic Test
    - Transitive
    - Transitive, across GOPATHs + collapse vendor/ directories.

Should Fix #358

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,16 @@
+# v42 2015/12/22
+
+* Fix a bunch of GO15VENDOREXPERIMENT issues
+    * Find package directories better. Previously we used build.FindOnly which didn't work the way I expected it to (any dir would work w/o error).
+    * Set the VendorExperiment bool based on go version as 1.6 defaults to on.
+    * A bunch of extra debugging for use while sanity checking myself.
+    * vendor flag for test structs.
+    * Some tests for vendor/ stuff:
+        * Basic Test
+        * Transitive
+        * Transitive, across GOPATHs + collapse vendor/ directories.
+* Should Fix #358
+
 # v41 2015/12/17
 
 * Don't rewrite packages outside of the project. This would happen if you specified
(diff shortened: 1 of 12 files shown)


freeformz:
@peter-edge @thockin This fixes the issues I know of and has test for the vendor/ related save issues. I also ran it against a slightly modified copy of `github.com/peter-edge/go-importserver` and `github.com/libopenstorage/openstorage` and the diffs, while not 100% clean look to be legit (although in the later case I only spot checked a of the differences).

thockin:
`godep save` now fails for me - specifically I have go-1.4 on this machine, but I am setting GO15VENDOREXPERIMENT=1 so that I can make a repo that works best in go-1.5, but still works with go-1.4.  godep v43 runs `godep save ./...` for a few minutes, then exits 0, but has not made any changes to the repo at all.  adding `-v` prints nothing at all.

PR #384, opened 2015-12-25
Ignore .go files when checking for license/legal

jonboulle:
The current heuristic for license or legal files is a little broad
because it will inadvertently sweep up go source files that shouldn't be
vendored. As one example, when trying to vendor the package
`github.com/camlistore/go4/errorutil`, `godep save` also brings in the
file `github.com/camlistore/go4/blob/master/legal/legal.go`.

This patch changes the `IsLicenseFile` and `IsLegalFile` checks to
immediately return false if the filename ends in `.go`.


freeformz:
This looks good, but what happens when a legal file get's dropped at `github.com/camlistore/go4/legal/LICENSE.md` ? AFAICT we'd include that one too when we probably shouldn't as errorutil doesn't depend on legal and shouldn't vendor anything from it.

jonboulle:
@freeformz indeed that is also a problem today ([e.g.](https://github.com/yifan-gu/kubernetes/commit/1693b740061a3ba78857b371b61e9197abf829be), but it's unrelated to this one - can we land this and fix that in a follow up?

freeformz:
@jonboulle I'm suggesting that fixing the later will also fix this issue, but in a more complete fashion.

client9:
Hi @jonboulle   I wrote the original code here.  My apologies that it's causing you troubles.

Just so I understand... is the problem just that godep just vendoring files it doesn't need to?  Or is this causing some other issue for you as well.

Another way of doing this is adding a HasSuffix check of `.txt`, `.rst`, `.md`, `.rtf`, will cover 99.999% of license files out there (well I don't know exactly but my previous scanning suggested this covered most of them).

Regards,

n

freeformz:

PR #387, opened 2015-12-29
Add Go devel version workaround

zchee:
Signed-off-by: Koichi Shiraishi <zchee.io@gmail.com>

### DIFF
--- a/dep.go
+++ b/dep.go
@@ -70,6 +70,9 @@ func uniq(a []string) []string {
 
 // trimGoVersion and return the major version
 func trimGoVersion(version string) (string, error) {
+	if version == "devel" {
+		return "devel", nil
+	}
 	p := strings.Split(version, ".")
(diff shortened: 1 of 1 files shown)


barakmich:
+1 for this, or a similar fix. My Travis builds are failing because of this (testing against Go `tip`).

freeformz:
Do you think we should record the sha?

```console
$ go version
go version devel +8d5ff2e Tue Dec 29 21:16:59 2015 +0000 darwin/amd64
```

Probably not worth the hassle. If you are using devel it will just always vary most likely.

freeformz:
Thanks!

zchee:
@freeformz Thanks merged.

> Do you think we should record the sha?
Yes. Since Golang recent change of cgo often, I also want sha.
Because I wanted to run test of travis, First it was pr only in 'devel'

PR #389, opened 2016-01-07
save: add a new option for skipping packages required by test files

mitake:
Current godep save tries to save packages which are required by test
related files in gopath. If test files are not required (-t isn't
passed), these packages are not needed to be stored under Godeps. This
commit lets save command not to save such packages if a new option -T
is passed.

### DIFF
--- a/diff.go
+++ b/diff.go
@@ -36,7 +36,7 @@ func runDiff(cmd *Command, args []string) {
 		GoVersion:  gold.GoVersion,
 	}
 
-	err = gnew.fill(dot, dot[0].ImportPath)
+	err = gnew.fill(dot, dot[0].ImportPath, true)
 	if err != nil {
 		log.Fatalln(err)
 	}
--- a/save.go
+++ b/save.go
@@ -19,7 +19,7 @@ import (
 
 var cmdSave = &Command{
 	Name:  "save",
-	Args:  "[-r] [-t] [packages]",
+	Args:  "[-r] [-t] [-T] [packages]",
 	Short: "list and copy dependencies into Godeps",
 	Long: `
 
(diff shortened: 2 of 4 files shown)


freeformz:
It would have to be a separate flag, or a modification to the -t flag (make it have different modes beyond true/false). Many people want to save the deps for their tests, but not the deps for their dep's tests.

Also, would require a test case in save_test.go

mitake:
@freeformz thanks for your review. I'll add a new flag for the purpose and update this PR later.

mitake:
@freeformz updated, the new PR adds a new flag '-T' to godep. With the option, godep skips packages required by test files (-T must not be used with -t at the same time). I also added a test case for the option.

mitake:
@freeformz sorry, I forgot to add the change for test :( I'll update this PR later.

mitake:
@freeformz updated, could you take a look?

save.go
```diff
@@ -56,19 +56,20 @@ files outside the project.
...
 var (
-	saveR, saveT bool
+	saveR, saveT, skipTestImports bool
 )
 
 func init() {
 	cmdSave.Flag.BoolVar(&saveR, "r", false, "rewrite import paths")
 	cmdSave.Flag.BoolVar(&saveT, "t", false, "save test files")
-
+	cmdSave.Flag.BoolVar(&skipTestImports, "T", false, "skip files that are imported by test files (it shouldn't be used with -t)")
```
freeformz: It's really "Don't process dependencies import by test files"

Also `it shouldn't be used with -t` conflicts with message above (`-T and -t must be given at the same time`).

diff.go
```diff
@@ -36,7 +36,7 @@ func runDiff(cmd *Command, args []string) {
 		GoVersion:  gold.GoVersion,
 	}
 
-	err = gnew.fill(dot, dot[0].ImportPath)
+	err = gnew.fill(dot, dot[0].ImportPath, true)
```
freeformz: What if this was saved with -T ?

freeformz:

PR #394, opened 2016-01-11
Added support for go development version vendoring.

drekar:
When go is compiled from source off go tip (master) and not a tagged version, the reported
major version is 'devel' rather than the released version number.

Thus a third conditional has been added to support this case when determining if vendoring should be enabled.

### DIFF
--- a/main.go
+++ b/main.go
@@ -89,7 +89,8 @@ func main() {
 	// VendorExperiment is the Go 1.5 vendor directory experiment flag, see
 	// https://github.com/golang/go/commit/183cc0cd41f06f83cb7a2490a499e3f9101befff
 	VendorExperiment = (majorGoVersion == "go1.5" && os.Getenv("GO15VENDOREXPERIMENT") == "1") ||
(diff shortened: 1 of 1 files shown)


freeformz:
Should devel not have the same rule as 1.6  and be written as:
```go
((majorGoVersion == "go1.6" || majorGoVersion == "devel" )&& os.Getenv("GO15VENDOREXPERIMENT") != "0")
```

drekar:
That works as well.  I was considering "devel" as a completely separate case from released version 1.6 for clarity and potentially divergent behavior from devel and the released, tagged version 1.6.  Future versions of 'devel' may not use the environment variable at all.

I can make that change.

freeformz:
Yes, the environment variable is supposed to be removed in go1.7 (so likely sometime after go1.6 release for devel). I'd say keep it split out, but fix the logic so it's inline with go1.6.

drekar:
That should do 'er then.  Thanks!

freeformz:
Thanks. This was merged as b922ec76f7137e4a33a07663f4cf8f9614e494cd.

PR #397, opened 2016-01-13
Document tag behavior

### LINKED ISSUE #391: v33-onwards: Undocumented change regarding including non-standard build tag dependencies on save.
Hi,

I've done a binary search on godep versions and traced this change of behavior/regression to version 33. It affects all versions starting from 33 onwards. Version 32 and earlier don't have this behavior.

The change is that now when doing `godep save`, it saves all dependencies even for non-standard build tags. This was previously not the case.

Here's a minimal reproduce case. Consider a Go package with 3 .go files:

```Go
package main

import (
	"fmt"
	"os"
)

func main() {
	_, err := assets.Open("foo.txt")
	fmt.Println(os.IsNotExist(err))
}
```

```Go
// +build !dev

package main

import "net/http"

var assets = http.Dir(".")
```

```Go
// +build dev

package main

import (
	"net/http"

	"github.com/shurcooL/httpfs/union"
)

var assets = union.New(map[string]http.FileSystem{
	"/assets": http.Dir("."),
})
```

With godep version 32, if you run `godep save ./...`, you'd get:

```JSON
{
	"ImportPath": "github.com/shurcooL/play/172",
	"GoVersion": "go1.5.2",
	"Packages": [
		"./..."
	],
	"Deps": []
}
```

Which makes sense because the normal package has no external dependencies (it only imports Go standard library).

But with version 33 and later, doing the exact same `godep save ./...` command, you get:

```JSON
{
	"ImportPath": "github.com/shurcooL/play/172",
	"GoVersion": "go1.5.2",
	"Packages": [
		"./..."
	],
	"Deps": [
		{
			"ImportPath": "github.com/shurcooL/httpfs/union",
			"Rev": "5782901d7de3a2107e252a57f5b3403213e7ec4f"
		},
		{
			"ImportPath": "github.com/shurcooL/httpfs/vfsutil",
			"Rev": "5782901d7de3a2107e252a57f5b3403213e7ec4f"
		},
		{
			"ImportPath": "golang.org/x/tools/godoc/vfs",
			"Rev": "4ad533583d0194672e7d3bc6fb8b67c8e905d853"
		}
	]
}
```

Note that it vendors dependencies that are only there if you specify `-tags=dev` build tag. They're not dependencies when doing a normal `godep go build`.

Is this a bug/regression or working as intended?

Given that it's a significant change of behavior that is not documented in the changelog, I'm tempted to understand it's an unintended regression.

This has a significant effect on packages that use build tags for some non-critical functionality (for example, a "dev" build tag for development mode, or "ignore" build tag for some go generate helper.

It also means doing `godep save ./...` on existing projects after 33 will potentially create a significant diff by adding more packages. If this behavior is meant to be, I want to confirm so that it doesn't unexpected change again without notice in a future version of godeps, causing yet another disruptive large unexpected diff when doing `godep save ./...` in an existing project.

Thanks.

freeformz:
Starting with v33 we started processing all go files, except those with
the `ignore` build tag. In v34 we also ignore files with the `appengine`
build tag.

Fixes #391

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -86,10 +86,12 @@
     * #350: undefined: build.MultiplePackageError
     * #351: stow away helper files
     * #353: cannot find package "appengine"
+        * Don't process imports of `.go` files tagged with the `appengine` build tag.
 
 # v33 2015/12/07
 
 * Replace the use of `go list`. This is a large change although all existing tests pass.
+    * Don't process the imports of `.go` files with the `ignore` build tag.
 
 # v32 2015/12/02
 
--- a/Readme.md
+++ b/Readme.md
@@ -52,6 +52,9 @@ command, you wrap it in one of these two ways:
 - When using a different command, set your `$GOPATH` using `godep path` as
   described below.
 
+Godep does not process the imports of `.go` files with either the `ignore` 
+or `appengine` build tags.
+
 Test files and testdata directories can be saved by adding `-t`.
 
 ## Additional Operations


freeformz:
@shurcooL Is this satisfactory for now?

Changelog.md
```diff
@@ -86,10 +86,12 @@
     * #350: undefined: build.MultiplePackageError
     * #351: stow away helper files
     * #353: cannot find package "appengine"
+        * Ignore files with the `appengine` build tag
 
 # v33 2015/12/07
 
 * Replace the use of `go list`. This is a large change although all existing tests pass.
+    * Ignore files with the `ignore` build tag. All other build tags are ignored.
```
shurcooL: The phrasing is a little confusing. Ignore is used with 2 different meanings.

How about:

> Ignore files with the `ignore` build tag. All other build tags are included and vendored.

Readme.md
```diff
@@ -52,6 +52,9 @@ command, you wrap it in one of these two ways:
 - When using a different command, set your `$GOPATH` using `godep path` as
   described below.
 
+Godep ignores files with either the `ignore` or `appengine` tags (does not process
+the deps in those files).
```
shurcooL: "does not vendor" maybe?

shurcooL:
Yeah, this helps a lot. If possible, tweak the language to be a little better and less confusing, but overall LGTM.

Readme.md
freeformz: it will actually vendor the files, but doesn't process the imports from them.

shurcooL: Right. :+1: It does not vendor the dependencies of those files.

Issue #409, opened 2016-02-04
`godep save` hangs when pwd is a symlink into $GOPATH


Manbeardo:
On OS X

Steps to reproduce:
* `go get github.com/my/sweet/project`
* `ln -s $GOPATH/src/github.com/my/sweet/project ~/project`
* `cd ~/project`
* `godep save ./...`
* wait for the heat death of the universe

`godep` works fine if `pwd` is an absolute path, but symlinks make it unhappy. This is problematic because symlinking a project dir from `~` is a pretty standard workflow to make switching projects easier. There's nothing unusual about my `$GOPATH` structure, just the way I `cd` into it from `~`.

freeformz:
godep doesn't support this. It may have inadvertently done so in the past however. I'm going to keep this open though for now while I reconsider.

alexflint:
I just ran into this problem, too.

freeformz:
Can you try a build of godep with #418 applied and see if that fixes things?

Manbeardo:
Still seeing the problem after applying #418. Double-check that my process is good?

- `cd $GOPATH/src/github.com/tools/godep`
- `git pull`
- `git cherry-pick 81f8a9945e94471bea4f838423494d84dd8ffc5b`
- `go install github.com/tools/godep`
- rerun godep

Manbeardo:
FWIW, resolving symlinks on Darwin is surprisingly painful.

freeformz:
@Manbeardo thanks for the update. I'll have to make a test case (even if it's manual) and figure out where else we fall down.

s7anley:
In my case the command is finished, but results is still bad. `ImportPath` is changed from project name to `.` and all dependencies are removed. As workaround I work directly from 
`$GOPATH/src/company-name/project`

freeformz:
Thanks for the test & report. I'll have to investigate more and see if I can make a test case.

aeden:
Same problem here. Workaround is simple enough, just cd to the actual location under $GOPATH and run `godep save` there.

gannulka:
Thank you ! This workaround saved me

freeformz:
FWIW: godep (v60) no prints out a warning whenever you are NOT in the $GOPATH.

freeformz:
Closing issues before archiving. Thank You\!

PR #418, opened 2016-02-23
Don't use git rev-parse --show-toplevel

freeformz:
`git rev-parse --show-toplevel` resolved symlinks, so don't use it. This
has caused soooo many problems. :-(

So instead we'll use `git rev-parse --show-cdup`, which outputs relative
paths to the root or an empty string. This wrapped in an Abs() should
get us what we want.


freeformz:
May (partially) fix #409.

freeformz:
May (partially) fix #334

PR #430, opened 2016-03-09
Add godep version to Godeps.json file

### LINKED ISSUE #429: Proposal: Add godep version to Godeps.json
A very common bug we've experienced is developers using an old version of godep when attempting to use this tool. Our first recommendation when devs have issues with godep is "are you on the latest version?" Additionally different versions of godep handle tests, licenses, etc differently. 

I propose adding the godep version to the Godeps.json file. At the very least, during code review we could see that a user has used an outdated version of godep to update vendored packages.

```json
{
	"ImportPath": "github.com/foo/bar",
	"GoVersion": "go1.6",
	"GodepVersion": "v53",
	"Packages": [
		"./..."
	],
	"Deps": [
	]
}
```

ericchiang:
As discussed in #429 this change adds the current godep version to Godeps.json when writing that file.

Closes #429

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -109,10 +109,11 @@ Godeps is a json file with the following structure:
 
 ```go
 type Godeps struct {
-	ImportPath string
-	GoVersion  string   // Abridged output of 'go version'.
-	Packages   []string // Arguments to godep save, if any.
-	Deps       []struct {
+	ImportPath   string
+	GoVersion    string   // Abridged output of 'go version'.
+	GodepVersion string   // Abridged output of 'godep version'
+	Packages     []string // Arguments to godep save, if any.
+	Deps         []struct {
 		ImportPath string
 		Comment    string // Description of commit, if present.
 		Rev        string // VCS-specific commit ID.
--- a/diff_test.go
+++ b/diff_test.go
@@ -1,17 +1,19 @@
 package main
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 )
 
-const (
-	d1 = `--- Godeps/Godeps.json
+var (
+	d1 = fmt.Sprintf(`--- Godeps/Godeps.json
 +++ $GOPATH
-@@ -1,12 +1,12 @@
+@@ -1,13 +1,13 @@
  {
         "ImportPath": "C",
         "GoVersion": "go1.2",
+        "GodepVersion": "v%d",
         "Deps": [
                 {
                         "ImportPath": "D101",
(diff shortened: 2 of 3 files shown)


ericchiang:
bump @freeformz (https://github.com/google/cadvisor/pull/1162#issuecomment-197068631)

freeformz:
So my main concern with this is Godep.json churn. But that's probably un-avoidable if we do want to record the version string. w/o using this version elsewhere I'm not sure this is all that useful. And I can see trying to enforce the version to add overhead to interactions with developers. But maybe that overhead is already being imposed, just in an ad-hoc way?

freeformz:
PS: I am interested in figuring this out, just thinking through the details here a little and the end goal.

ericchiang:
> So my main concern with this is Godep.json churn. But that's probably un-avoidable if we do want to record the version string.

The version would only be changed when you write to Godeps.json (`godep save` or `godep get`). So it's inherently going to be part of some larger diff.

> w/o using this version elsewhere I'm not sure this is all that useful.

Hard to use it until the version's already in Godeps.json :)

> And I can see trying to enforce the version to add overhead to interactions with developers. But maybe that overhead is already being imposed, just in an ad-hoc way?

Yeah that's definitely what I'm getting at. We already have to enforce versions with developers since godep has been around a while (I worked with someone with v20something installed yesterday).

But also it's really useful to know when you're running a more recent version of godep as well. When I run a `godep save` and the diff of the Godeps directory has a bunch of files removed that had nothing to do with the packages I was trying to alter, it's very hard to know if that's user error or because someone was using a really old version of Godeps for the last commit.

I can see the biggest annoyance begin when you get a commit when the version goes from "v62" to "v61". But again, this should only happen when you're updating a vendored package anyway and those diffs are generally pretty big.

freeformz:
`save` or `update` will modify Godeps.json. `godep get` just (currently, I have a notion to change this to something more useful given `vendor/`) shells out to go get with some wrapping. I'm not sure the version should be updated on `godep update` though, since it's not a total re-save, more like selective editing. So if you see a Godeps.json file that had a small modification done you can't really assume it has all of the package fixes from recent versions for instance.

I don't know if that's enough to re-consider this or not.

I guess looking at the git history and the comments would probably tell you enough though about what was done?

ericchiang:
> I guess looking at the git history and the comments would probably tell you enough though about what was done?

How so? Without the version in the Godeps.json file we can't tell what version of godep was used to generate which commit.

freeformz:
Err Sorry. I meant the commit log with the version numbers and comments like "Updates github.com/foo/bar" would tell you.

freeformz:
Thanks!

Issue #448, opened 2016-04-05
godep does not respect +build tags, latest golang.org/x/net/context breaks


themartorana:
The [latest commit](https://github.com/golang/net/commit/afc22ee9133171570a267463982b542fb3edff5a) to the `context` package contains the file [go17.go](https://github.com/golang/net/blob/afc22ee9133171570a267463982b542fb3edff5a/context/go17.go) which imports the `context` library from the Go 1.7 standard library. 

Despite having the `// +build go1.7` tag at the top of the file, godep parses the file and throws a `godep: Package (context) not found` error.

I can get past this by going into the context package and checking out the `release-branch.go1.6` branch, but this simply illustrates the fact that `godep` isn't checking the `+build` tags, and is attempting to import packages for non-installed versions of Go.

themartorana:
Sorry, tapped the wrong button...

CannibalVox:
It's worse than that- deleting go17.go wont' help because it actually refused to read pre_go17.go.  It's like build tag is reversed, it will ONLY read the 1.7 file.

freeformz:
What version of godep? Versions of godep (circa v41+, I would need to check the changelog for the specific version) process all .go files, except for those with "ignore" and "appengine" build tags. Adding go version build tags shouldn't be that hard since we also store the major go version in the Godeps.json file though.

themartorana:
For me: `godep v60 (darwin/amd64/go1.6)`

CannibalVox:
Above PR fixes the issue, hopefully we can see the resolution live soon, this is a pretty serious deal :\

freeformz:
Looks like go1.7 includes a package named "context" and if you are not on 1.7 (really devel) then godep can't find the package because you don't actually have it.

freeformz:
Can anyone test our the branch that I pushed for this? I tested it here, but it's late and I don't want to merge it before reviewing it in the morning (or getting other +1s).

ddgenome:
@freeformz I can confirm that using PR https://github.com/tools/godep/pull/451 `godep save -v` on a project that imports "golang.org/x/net/context" finishes successfully whereas it fails with `godep: Package (context) not found` when using master.

themartorana:
@freeformz seconded.

freeformz:
Thanks everyone, this should now be fixed.

PR #449, opened 2016-04-05
Respect goversion build directives.

CannibalVox:
Yesterday afternoon, context rolled out a pre-1.7 update that uses build directives to control whether the 1.6 or 1.7 files are compiled.  These are in the format +build go1.7 and +build !go1.7.  Godep does not currently respect these build directives and chokes trying to vendor context.  This should repair this behavior, allowing us to use context again.

### DIFF
--- a/list.go
+++ b/list.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"unicode"
@@ -216,6 +217,10 @@ func fillPackage(p *build.Package) error {
 	if p.Goroot {
 		return nil
 	}
+	version, err := goVersion()
+	if err != nil {
+		return err
+	}
 
 	if p.SrcRoot == "" {
 		for _, base := range build.Default.SrcDirs() {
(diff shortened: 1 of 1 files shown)


CannibalVox:
Related to #448

list.go
```diff
@@ -216,6 +217,10 @@ func fillPackage(p *build.Package) error {
 	if p.Goroot {
 		return nil
 	}
+	version, err := goVersion()
```
freeformz: I think this should use the recorded go version from the Godeps.json file, not whichever go version is currently running.

list.go
```diff
@@ -271,6 +276,18 @@ NextFile:
 							p.IgnoredGoFiles = append(p.IgnoredGoFiles, fname)
 							continue NextFile
 						}
+						matchesPositive, err := regexp.MatchString("go\\d+\\.\\d+", b)
```
freeformz: compile this regex at the package level in a variable and re-use maybe? allows this to be written nicer/shorter at least.

list.go
```diff
@@ -271,6 +276,18 @@ NextFile:
...
+						if err != nil {
+							return err
+						}
+						if matchesPositive && b != version {
```
freeformz: Version build tags match the declared build tag version and newer. eg: a file with the `go1.5` tag matches go1.5 and go1.6 and go1.7, etc.

list.go
```diff
@@ -271,6 +276,18 @@ NextFile:
...
+							p.IgnoredGoFiles = append(p.IgnoredGoFiles, fname)
+							continue NextFile
+						}
+						if b == ("!" + version) {
```
freeformz: I *think* negatives also work the same way. eg. a tag of `!go1.5` means NOT go1.5 OR go1.6 or go1.7, etc.

freeformz:
Closing in favor of #451 for now.

Thanks for taking the time to take a stab at this though.

PR #451, opened 2016-04-06
Handle go version tags

### LINKED ISSUE #448: godep does not respect +build tags, latest golang.org/x/net/context breaks

freeformz:
Fixes #448

Obey the go tool's definition of how the go versions build tags work for
the purposes of dependency scanning.

Don't use the current go version, but the recorded go version in
Godeps.json.

FIXME: We still copy the files. This is sub optimal, but probably fine
in the 80% case where people are using the version of go they have
specified in the Godeps.json file. Longer term we need to obey the
gofiles / ignore lists from the package list.

IMO, Supersedes #449

### DIFF
--- a/list.go
+++ b/list.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"unicode"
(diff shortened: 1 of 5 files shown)


ddgenome:
This fixes issue #448 for me.

freeformz:
Merged as 35ee059b4e6c3df99fb6afc2d3b0f0790b22e1c1

PR #454, opened 2016-04-13
Support `tip` as `GoVersion`

jingweno:
/cc @freeformz


jingweno:
Closing after discussion. Tip is not a thing when Go reporting version.

Issue #453, opened 2016-04-07
godep restore makes impossible to update project with go get


Raffo:
Currently ```godep restore``` puts the repositories in detached HEAD and this makes impossible to execute a ```go get -u github.com/foo/bar```. If a high number of projects using godeps are used and if godep restore is used frequently, this leads to a very high number of projects in detached state which makes very hard to update them. 
Are we able to avoid this somehow?

freeformz:
Longer term, yes, by removing the need to `godep restore`. A go tooling change recently (go1.6) made this a problem AFAIK.

freeformz:
PS: godep restore is doing what it's always done, restore the recorded versions to the $GOPATH. it can only do this by checking out the various shas (AFAIK) it's recorded.

As a temp work around I could implement a `godep unrestore`, which would check out master again. But I'm hesitant to add another command that should be removed in the future and I'm not sure it's super helpful in the workflow ... ie. you have to remember to use it.

Thoughts?

Raffo:
I was about to implement a ```godep unrestore``` myself, I think it would be useful (I can work on it and submit a PR, but if you want to work on it, as you whish). More important, though is to document in the ```README``` what happens now if a user executes ```godep restore```. 

Also, can you give me the detail of this change in 1.6 you are mentioning above?

freeformz:
PRs welcome. I don't have an exact commit to golang where this broke. Let me see if I can dig it up though.

freeformz:
It was likely this one: https://github.com/golang/go/commit/42206598671a44111c8f726ad33dc7b265bdf669

freeformz:
@Raffo What would you suggest be added to the README? It states what `godep restore` already does in a non VCS specific manner, since godep handles multiple types of VCS.

Raffo:
Thanks for the link to the commit. 

Do you think that a warning like "```godep restore``` could leave some of the dependencies in a detached state. This could lead to problems to update your dependencies" would be too much? Maybe we could add some text to specify that the users should avoid using ```godep restore``` if not strictly necessary.

freeformz:
How about

**NOTE**: godep restore leaves git repositories in a detached state. `go1.6`+ no longer checks out the master branch when doing a `go get`, see [here](https://github.com/golang/go/commit/42206598671a44111c8f726ad33dc7b265bdf669).

Then if you land an `unrestore` PR we can adjust that to recommend the usage of `unrestore` afterwards?

Raffo:
[...]

PR #464, opened 2016-05-10
handle detached git on restore

freeformz:
If the git repo is detached, we need to determine the default remote
branch before attempting to donwload (can't pull into a repo in detached
state).

This attempts to use "git remote show origin" to have the origin tell us
what the default branch is.

PS: this may not work with old (~3 year old git servers).


freeformz:
This is because of this: https://github.com/golang/go/commit/42206598671a44111c8f726ad33dc7b265bdf669

This doesn't "fix" anything for non godep users, just allows godep restore to work properly.

freeformz:
This came up in #453

Issue #465, opened 2016-05-12
update a single package


yields:
```bash
$ godep update github.com/pkg/a 
godep: Package (github.com/pkg/b) not found
```

I thought that `godep update <pkg>` will update the given package, but seems like it "checks" all dependencies and errors if a dependency is not found, even though `github.com/pkg/a` has no dependencies.

Any reason to check them all, is there a workaround to updating a single package?

--

```
$ godep version
godep: WARNING: Go version (go1.6) & $GO15VENDOREXPERIMENT= wants to enable the vendor experiment, but disabling because a Godep workspace (Godeps/_workspace) exists
godep v66 (darwin/amd64/go1.6.1)
```

freeformz:
The workflow for godep update is to first `godep restore`, then checkout the version of the package you want to update to, then update.

yields:
I see, is there a workaround at all?

Seems to me that `godep <pkg>` should update a single package and it's dependencies and not error if another package does not exists, is there a reason for this workflow that i'm missing?

Would love to work on a fix if you're interested :)

freeformz:
I agree. I thought there was a test for this, but maybe not.

PR #467, opened 2016-05-12
Handle missing deps a little better

freeformz:
I added 2 tests.

The first one covers the case where we're updating a
package/... that has removed some dependencies.

The second one covers the case where we're update a package, and
another, non transitive package is missing from $GOPATH.

I'm not sure this covers everything. I need to re-read the other update
tests (but ATM they pass).

### DIFF
--- a/update_test.go
+++ b/update_test.go
@@ -528,7 +528,7 @@ func TestUpdate(t *testing.T) {
 				},
 			},
 		},
-		{ // 11 - package/..., new transitive package, different repo
+		{ // 12 - package/..., new transitive package, different repo
 			vendor: true,
 			cwd:    "C",
 			args:   []string{"D/..."},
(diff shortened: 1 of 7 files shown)


update_test.go
```diff
@@ -579,6 +579,106 @@ func TestUpdate(t *testing.T) {
...
+				{"C/vendor/D/E/main.go", "(absent)", nil},
+			},
+			wdep: Godeps{
+				ImportPath: "C",
+				Deps: []Dependency{
+					{ImportPath: "D/A", Comment: "D2"},
+				},
+			},
+		},
+		{ // 14 - Update package A, but not package B, which is missing from $GOPATH
```
freeformz: @yields This works, at least on this branch. Does this demonstrate what you were talking about in #465?

yields: Thanks for the quick fix!

I've just tried this locally, it updates the pkg/a correctly, but there are two small UX issues:

- it still outputs an error that `pkg/b` was not found
- it doesn't work for `pkg/a/sub`, so you have to do `pkg/a/...`

Anything i can do to help get this merged? :)

freeformz: merging now. I'll look at cleaning up UX next week. That may actually be harder. ;-)

Issue #377, opened 2015-12-22
godep: Unable to find dependent package golang.org/x/sys/unix in context of /Users/apple/workspace/go/src/github.com/Sirupsen/logrus


ghost:
on go 1.5.2
I cant work out why it screams for logrus.

```
gedw99-MacBook-Pro:backend apple$ go env
GOARCH="amd64"
GOBIN="/Users/apple/workspace/go/bin"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/apple/workspace/go"
GORACE=""
GOROOT="/usr/local/go"
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
GO15VENDOREXPERIMENT=""
CC="clang"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fno-common"
CXX="clang++"
CGO_ENABLED="1"



gedw99-MacBook-Pro:backend apple$ godep save -r
godep: Unable to find dependent package golang.org/x/sys/unix in context of /Users/apple/workspace/go/src/github.com/Sirupsen/logrus
gedw99-MacBook-Pro:backend apple$ go get github.com/Sirupsen/logrus
gedw99-MacBook-Pro:backend apple$ godep save -r
godep: Unable to find dependent package golang.org/x/sys/unix in context of /Users/apple/workspace/go/src/github.com/Sirupsen/logrus
gedw99-MacBook-Pro:backend apple$ 


```

gedw99:
This is a bug i think.
Once i did a go get golang.org/x/sys/unix, then godep was happy.
So it should be included in godep binary or whatever ? 

````

dw99-MacBook-Pro:backend apple$ godep save -r
godep: Unable to find dependent package golang.org/x/sys/unix in context of /Users/apple/workspace/go/src/github.com/Sirupsen/logrus
gedw99-MacBook-Pro:backend apple$ go get golang.org/x/sys/unix
gedw99-MacBook-Pro:backend apple$ godep save -r
gedw99-MacBook-Pro:backend apple$ 


```

freeformz:
It is a transitive dependency that "go get" does not fetch, but can be required on other platforms to build the package. Older versions of godep relied on the go tool, but we no longer do, for exactly this reason.

I plan on enhancing this workflow, especially around GO15VENDOREXPERIMENT.

[...]

PR #473, opened 2016-05-18
Add -b option to follow build constraints

roxma:
See this issue: https://github.com/tools/godep/issues/377#issuecomment-219917019(https://github.com/tools/godep/issues/377#issuecomment-219917019)


freeformz:
I'd much rather do this by making [this](https://github.com/tools/godep/blob/master/list.go#L22) configurable via an array in Godps.json. Plus ensure that our code handles all the various permutations of build tags. And maybe default to what is hard coded now.

PR #479, opened 2016-05-29
'determing' to 'determining'

harryge00:
I guess this is a typo?


freeformz:
It is indeed. Thanks!

Issue #481, opened 2016-05-31
godep save fails on windows with "chmod not supported by windows error"


alexbrand:
### Expected behavior

Godep should copy the dependencies into the Godeps file and vendor folder
### Actual behavior

Error is shown saying that chmod is not supported by Windows. Sample:
`godep: chmod vendor\github.com\square\go-jose\json\decode.go: not supported by windows`
### Steps to reproduce behavior

Run `godep save ./...` on a new project that has dependencies.
### `godep version` output

`godep v72 (windows/amd64/go1.6.1)`
### `go version` output

`go version go1.6.1 windows/amd64`

freeformz:
Closing issues before archiving. Thank You\!

PR #482, opened 2016-05-31
Use os.Chmod instead of file.Chmod for compatibility with Windows

alexbrand:
Address #481 by using os.Chmod instead of file.Chmod.


freeformz:
Thanks!

PR #499, opened 2016-07-22
Add foreach command to run a shell command in each dependency's directory

andrewstuart:
I think this gives a very simple solution to #453 of simply running `godep foreach git checkout master` to fix any inconsistencies caused by `godep restore`. I'm sure it could be useful in other ways as well, and matches up well with the `git submodule foreach` paradigm.

I'm sure there are other enhancements that could be made to this, so I'd be glad to go through the work to get it up to the necessary quality for inclusion. Let me know what I can do. :smile:

### DIFF
--- a/foreach.go
+++ b/foreach.go
@@ -1,0 +1,51 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"os/exec"
+)
+
+var cmdForeach = &Command{
+	Name:         "foreach",
+	Short:        "run a command in each dependency's path",
+	Run:          runForeach,
+	OnlyInGOPATH: true,
+}
+
+func runForeach(cmd *Command, args []string) {
+	g, err := loadDefaultGodepsFile()
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	if args == nil || len(args) < 1 {
+		log.Fatalln("Must provide a command to foreach")
+	}
+
+	for _, dep := range g.Deps {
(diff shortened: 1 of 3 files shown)


freeformz:
Blinly checking out master isn't a good thing, but probably works in the 90% case. I'd argue that the appropriate place to fix this is to have `go get` determine the remote head and check that out when not already on that branch when doing a `go get -u`.  godep already has logic to do that because it needs to check out the correct default head when updating to ensure the correct sha is present.

Also my goal, although I don't know if I'll ever get to it, is to remove the need to ever godep restore.

Thoughts?

andrewstuart:
Yeah, it'd definitely be nice to not need to run `godep restore`, but I still think a generic `foreach` command has the potential to be useful for one-offs and scripts, much as it is with `git submodule foreach`.

I think it's just a nice side-effect that for now it does help the process of upgrading old Godeps/_workspace environments using `godep restore; godep save; godep foreach git checkout master`

foreach.go
```diff
@@ -0,0 +1,57 @@
...
+		log.Fatalln(err)
+	}
+
+	if args == nil || len(args) < 1 {
+		log.Fatalln("Must provide a command to foreach")
+	}
+
+	for _, dep := range g.Deps {
+		c := exec.Command(args[0], args[1:]...)
+		c.Dir = getRoot(dep.root, dep.ImportPath)
```
freeformz: Should this be run for each dep or each repo?

PR #503, opened 2016-08-02
option to exclude packages that match a pattern

asquare:
add the ability to exclude packages that match a pattern


freeformz:

PR #518, opened 2016-10-02
it needs some modules

shinriyo:
added some installing some modules before install


parkr:
These should be automatically installed when you run `go get github.com/tools/godep`. What error did you see when these are not installed?

freeformz:

PR #524, opened 2016-12-03
add support for go1.8

tobstarr:
In order to test go 1.8 as soon as possible we should be able to also use godep already.

I could also submit a more advanced fix (which would also cover go version 1.9, 1.10, etc.) but I am not sure if that is wanted?


tobstarr:
ping @freeformz

tobstarr:
👍

dlsniper:
@freeformz can you please push the button to merge && release this. Thank you!

PR #528, opened 2017-01-13
Don't include quotes around hg revisions

euank:
Prior to this change, the saved revision in a Godeps.json would include
the single quotes around it from the template.

An example of what the mistake this fixes looks like in the wild: https://github.com/kubernetes/kubernetes/blob/v1.5.1/Godeps/Godeps.json#L16-L20


freeformz:
Thanks!

PR #530, opened 2017-01-21
Bump version number to 78

moncho:
I guess this change was forgotten before releasing v78.

### DIFF
--- a/version.go
+++ b/version.go
@@ -8,7 +8,7 @@ import (
 	"strings"
 )
 
-const version = 77
+const version = 78
 
 var cmdVersion = &Command{
 	Name:  "version",


freeformz:
Ugh. I did. I really need to write a `make release`.

Thanks!

MikeMcQuaid:
A gentle request: please don't retag things, folks. If you use Git you'll not receive the updated tag without manually deleting and refetching it as it's considered rewriting public history. It also means you break the checksums for things like Homebrew that rely on your tags (https://github.com/Homebrew/homebrew-core/issues/9293). Thanks!

freeformz:
Yeah. sorry, although I won't promise to not do that again. IMO tags are mutable and should be treated as such, although I do agree that doing so should be discouraged.

MikeMcQuaid:
Thanks! 

> IMO tags are mutable and should be treated as such

Just FYI: Git disagrees 😉

MikeMcQuaid:
(well, as much as Git considers anything immutable 😆)

Thanks again!

PR #538, opened 2017-04-06
Skip download if ImportPath prefix has been seen

sttts:
The restore code looks up the repo root via `vcs.RepoRootForImportPath` for each imported package. If there are many (literally hundreds in our case) sub-packages of the same dependency, this code looks up the repo root again and again. If the dependency is on a non-known domain (e.g. `k8s.io/kubernetes` in contrast to `github.com/foo/bar`), this leads to one `go get` per sub-package, i.e. one http connection to get the meta data per sub-package. This is slow, very slow.

### DIFF
--- a/restore.go
+++ b/restore.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"golang.org/x/tools/go/vcs"
 )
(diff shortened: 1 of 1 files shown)


ericchiang:
FYI this change also dramatically reduces the memory usage of `godep restore` (~3 GB) when using it against the Kubernetes repo.

philips:
cc @thockin

thockin:
Good to know, but I won't have time to futz with glide any time soon.

freeformz:
I am closing this PR because the repo is going to be archived.

